PR #6, opened 2026-05-26
Add folder picker UI and dir-browse API

PBnicad:
Introduce a server API to browse filesystem directories and a client FolderPicker component, and wire it into SessionSidebar. Adds app/api/dir-browse/route.ts to list directory entries (filters common ignored names, supports Windows drive discovery, home/default handling, parent navigation and error responses). Adds components/FolderPicker.tsx — a client-side modal for drive selection, path input, browsing and selecting folders. Integrates the picker into components/SessionSidebar.tsx by adding a browse button next to the custom path input and opening the FolderPicker to populate the path. Includes basic UI/UX behaviors (loading, errors, keyboard handling) and focused styling adjustments.

### DIFF
--- a/app/api/dir-browse/route.ts
+++ b/app/api/dir-browse/route.ts
@@ -1,0 +1,103 @@
+import { NextRequest, NextResponse } from "next/server";
+import fs from "fs";
+import path from "path";
+import os from "os";
+
+const IGNORED_NAMES = new Set([
+  "node_modules", ".git", ".next", "dist", "build", "__pycache__",
+  ".turbo", ".cache", "coverage", ".pytest_cache", ".mypy_cache",
+  "target", "vendor", ".DS_Store",
+]);
+
+interface DirEntry {
+  name: string;
+  isDir: boolean;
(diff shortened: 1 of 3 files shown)


agegr:
There’s a flow issue with the folder picker selection.

`FolderPicker` is rendered outside `dropdownRef`, so clicking the picker’s Select button first triggers the sidebar’s document-level `mousedown` outside-click handler. That closes the cwd dropdown, closes the custom path editor, and clears `customPathValue`.

After that, the Select button’s `onClick` only calls:

```tsx
onSelect={(path) => { setCustomPathValue(path); setFolderPickerOpen(false); }}
```

So the selected path is written back into state, but the custom path UI is already hidden and the cwd is not actually changed. From the user’s perspective, selecting a folder just closes everything without opening/switching to that folder.

I think selecting a folder should either directly call the same logic as “Open” / update `selectedCwd`, or the outside-click handler should ignore clicks inside the folder picker and keep the custom path controls visible.

PR #10, opened 2026-05-26
fix: expand tilde in file path for ~/pi-workspace paths

ayakabot:
### Problem

When a user sets a custom workspace path like `~/pi-workspace` in the UI, the FileExplorer fails to load with:

```
GET /api/files/root/pi-workspace?type=list
{"error":"Not found"}
```

The issue is that `filePathFromSegments()` in `app/api/files/[...path]/route.ts` does not expand the tilde (`~`) to the user's home directory. The path `~/pi-workspace` is passed as a literal string to `fs.statSync()`, which looks for a directory literally named `~` instead of `/home/user`.

### Root Cause

In `filePathFromSegments()`, the function joins URL-encoded path segments but never handles the tilde prefix. Since URLs encode `~` as `%7E`, the final path becomes `/%7E/pi-workspace` or similar, which doesn't exist.

### Solution

Added tilde expansion in `filePathFromSegments()`:

```typescript
if (slashJoined === "~" || slashJoined.startsWith("~/")) {
  return slashJoined.replace(/^~/, os.homedir());
}
```

This expands `~` and `~/...` paths to the user's actual home directory before performing filesystem operations.

### Testing

1. Create a new directory manually: `mkdir -p ~/pi-workspace`
2. Set the custom workspace path to `~/pi-workspace` in pi-web
3. Verify the FileExplorer loads correctly and shows the directory contents

### DIFF
--- a/app/api/files/[...path]/route.ts
+++ b/app/api/files/[...path]/route.ts
@@ -1,6 +1,7 @@
 import { NextRequest, NextResponse } from "next/server";
 import fs from "fs";
 import path from "path";
+import os from "os";
 import { listAllSessions } from "@/lib/session-reader";
 
 const IGNORED_NAMES = new Set([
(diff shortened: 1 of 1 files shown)


agegr:
我测了一下，只修复files还不太够。
review by codex
结论：这个 PR 的修复不完整。它只在 /api/files 的请求路径上展开了 ~，但 allow-list 里的 cwd 以及新建会话时的 existsSync(cwd) 仍然不会展开 ~。因此 ~/pi-workspace 场景下 FileExplorer 可能仍然被 403 拦截，新会话发送消息也可能失败。建议把 ~ 展开抽成统一的路径规范化逻辑，并在文件 API、allowed roots、新会话 cwd 校验中一起使用。

PR #26, opened 2026-05-28
feat: add command copy button to tool call blocks

kami1983:
Show a copy icon next to tool call headers when the tool input contains a command string
Add cwdExists flag to session list API and filter out sessions with deleted cwd in sidebar
Improve error handling in useAgentSession (clear previous errors, parse error response body)

### DIFF
--- a/app/api/sessions/route.ts
+++ b/app/api/sessions/route.ts
@@ -1,10 +1,16 @@
 import { NextResponse } from "next/server";
+import { existsSync } from "fs";
 import { listAllSessions } from "@/lib/session-reader";
 
 export async function GET() {
   try {
     const sessions = await listAllSessions();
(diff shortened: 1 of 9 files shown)


agegr:
- Missing cwd should not imply the session no longer exists. The session history is still valuable, so I think missing-cwd sessions should remain visible, preferably with a “missing directory” badge and disabled/limited File Explorer or send behavior.
- `useAgentSession` now stores send failures in the top-level `error`, which makes `ChatWindow` render the fatal error screen and removes the composer. Send errors should be non-fatal so the user can edit/retry.
- The new tool-input copy button is useful, but visually it should match the existing “Copy message” affordance and copied state for consistency.
- The print CSS is global and can make non-Markdown pages print blank. It should be scoped only when printing Markdown preview.

PR #38, opened 2026-05-31
fix: remove tree field from session detail API to prevent stack overflow on long linear sessions

fallleave001:
## Summary

Fix HTTP 500 `RangeError: Maximum call stack size exceeded` when loading a session with 2000+ messages that has never used the `/tree` branch feature.

## Root Cause

`GET /api/sessions/[id]` includes the `tree` field from `SessionManager.getTree()`. For a linear session with no branches, this produces a deeply nested chain where each level contains the next message. With 3523 messages, the tree is 3511 levels deep. `NextResponse.json()` → `JSON.stringify()` overflows Node.js's call stack at ~2000 levels.

## Fix

Remove `tree: m` from the response JSON. The `tree` field is not used by the UI — session list and message content display correctly without it.

## Testing

1. Create a long conversation (2000+ messages) without using `/tree` branches
2. Open the session in pi-web — should load normally (was HTTP 500)
3. Verify all messages are displayed correctly

### DIFF
--- a/app/api/sessions/[id]/route.ts
+++ b/app/api/sessions/[id]/route.ts
@@ -23,7 +23,6 @@ export async function GET(
 
     const sm = SessionManager.open(filePath);
     const entries = sm.getEntries() as never;
-    const tree = sm.getTree();
     const leafId = sm.getLeafId();
     const context = buildSessionContext(entries, leafId);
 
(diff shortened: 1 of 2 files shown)


agegr:
The stack overflow root cause looks valid, but the fix removes an API field that the UI still consumes for in-session branch navigation. `useAgentSession` passes `data?.tree ?? []` to `BranchNavigator`, so branched sessions will now appear to have no branches and users lose the ability to switch leaves.

The manual testing only covers long linear sessions. Please also test a session that has used `/tree`, or replace the recursive `tree` payload with a non-recursive / lazy branch representation instead of removing the field outright.

PR #40, opened 2026-05-31
fix: read pi and app versions at runtime instead of baking at build time

fallleave001:
NEXT_PUBLIC_* env vars are resolved at build time, so upgrading pi-coding-agent after building won't update the displayed version. Fix by reading versions via /api/version at runtime.


agegr:
不建议用户自己更新pi web的依赖

PR #42, opened 2026-05-31
fix: flatten session tree in API response to avoid JSON serialization…

LQFHUB:
… stack overflow

NextResponse.json() uses undici's internal JSON serializer which has a smaller recursion depth limit than V8's native JSON.stringify. When a session has a long linear entry chain (2200+ levels), the deeply nested tree causes RangeError: Maximum call stack size exceeded during serialization.

Fix: flatten the tree before serialization (children stored as string IDs instead of nested nodes). The frontend detects the flat format and rebuilds the nesting for components that require it (BranchNavigator).

Fixes #1

### DIFF
--- a/app/api/sessions/[id]/route.ts
+++ b/app/api/sessions/[id]/route.ts
@@ -62,11 +62,27 @@ export async function GET(
       }
     }
 
+    // Flatten tree to avoid undici's JSON serializer stack overflow on deeply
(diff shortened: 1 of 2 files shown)


agegr:
The fix addresses a real issue, but the current patch does not typecheck and also changes branch ordering in the rebuilt tree. Please fix those before merge.

PR #52, opened 2026-06-05
[codex] Add math rendering, cwd persistence, and audio attachments

Surfire:
## Summary

Adds four UI/runtime improvements:

- Render Markdown math in chat messages and Markdown file previews via `remark-math`, `rehype-katex`, and KaTeX CSS.
- Normalize and persist selected cwd values, including `~` and relative paths, so custom cwd selection survives refresh and new sessions receive absolute paths.
- Change the local default cwd endpoint to use `~/dev/pi-cwd` instead of date-stamped `~/pi-cwd-YYYYMMDD` directories.
- Extend chat attachments from images to images plus audio, forwarding audio media blocks through the existing multimodal prompt path for providers that accept inline audio data.

Also includes the small session tree compression fix from open PR #48 to avoid deep linear tree serialization issues, and lets the local linked `pi-web` binary fall back to `next dev` when no production build exists.

## Validation

- `npm run lint`
- Browser verified `http://localhost:30141` renders existing math as KaTeX nodes.
- Browser verified custom cwd `~/dev/pi-cwd` normalizes to `/Users/chenmin/dev/pi-cwd` and survives reload.
- Browser verified the file input accepts `image/*,audio/*` and the attach button is labeled for image or audio.
- Local API verified `POST /api/default-cwd` returns `/Users/chenmin/dev/pi-cwd`.

## Notes

The current `@earendil-works/pi-coding-agent` public types only expose image attachments. Audio is forwarded as an inline media block through the same prompt path, which should work for Google/Gemini-style providers that accept inline audio data, but it is not a universal provider capability.

### DIFF
--- a/app/api/agent/new/route.ts
+++ b/app/api/agent/new/route.ts
@@ -1,6 +1,7 @@
 import { NextResponse } from "next/server";
 import { existsSync } from "fs";
 import { startRpcSession } from "@/lib/rpc-manager";
+import { normalizeCwd } from "@/lib/cwd";
 
 // POST /api/agent/new  body: { cwd: string; type: string; message: string; ... }
 // Spawns a brand-new pi session and immediately sends the first command.
@@ -13,7 +14,8 @@ export async function POST(req: Request) {
     if (!cwd || typeof cwd !== "string") {
       return NextResponse.json({ error: "cwd is required" }, { status: 400 });
     }
-    if (!existsSync(cwd)) {
+    const normalizedCwd = normalizeCwd(cwd);
+    if (!existsSync(normalizedCwd)) {
       return NextResponse.json({ error: `Directory does not exist: ${cwd}` }, { status: 400 });
     }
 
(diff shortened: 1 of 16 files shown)


agegr:
Overall this is a useful direction, and the math rendering plus cwd normalization pieces look reasonable to me.

I don’t think the PR is ready to merge as-is though:

1. Audio attachments are only enabled in the UI. The backend still sends them through the existing `images` option and uses `as never` to bypass the SDK types. `pi-coding-agent` / `pi-ai` currently only define image attachments there, so providers may silently drop audio or mis-handle it as image data.

2. The session tree compression fix does not fully solve the deep linear tree issue. `compressTree()` is still recursive, so a sufficiently deep chain can still hit `Maximum call stack size exceeded`. It also compresses the API response even though the UI already has display-level compression, which risks losing useful tree structure.

I’d suggest splitting audio support out until there is a real SDK/provider contract for audio, and changing the tree fix to an iterative/display-level approach. The math and cwd changes seem fine to merge separately.

PR #65, opened 2026-06-08
feat: add scroll-to-bottom control for chat sessions

### LINKED ISSUE #59: Avoid Auto-Scrolling to Bottom After Response Completion
This is truly a simple, elegant, and efficient tool. It’s very nice to use. I hope that after an answer finishes, the page won’t automatically jump to the bottom immediately. If I start reading when only a small part has been generated, and then the response suddenly finishes, it automatically scrolls to the bottom, which isn’t ideal. It would be better to add a “scroll to bottom” button (for example, on the conversation line on the right side, or wherever you think is appropriate). Thank you very much.

oa1mgo:
## Summary

Fixes #59.

This PR updates the chat scroll behavior so the view no longer jumps to the bottom after a response completes when the user has scrolled away from the latest message.

It also keeps the chat pinned to the latest message when new or streaming content arrives while the user is already at the bottom.

It adds a lightweight scroll-to-latest control that appears when the user is away from the bottom of the conversation, and removes the previous running-state bottom spacer that pushed the message list upward while the agent was active.

## Changes

- Track whether the chat viewport is near the bottom before auto-scrolling
- Preserve the user's reading position while an agent response is streaming or completing
- Keep the chat pinned to the latest message when new or streaming content arrives while the user is already at the bottom
- Remove the running-state bottom spacer and last-user-message scroll anchoring
- Add a centered scroll-to-bottom button for returning to the latest message
- Hide the button naturally once the viewport reaches the bottom

## Testing

- `npm run lint`
- Manually verified the scroll button appears, scrolls smoothly to the latest message, does not flicker on click, and preserves bottom auto-follow for new messages

### DIFF
--- a/components/ChatWindow.tsx
+++ b/components/ChatWindow.tsx
@@ -96,13 +96,12 @@ export function ChatWindow({ session, newSessionCwd, onAgentEnd, onSessionCreate
     agentRunning, modelNames, modelList, modelThinkingLevels, modelThinkingLevelMaps, toolPreset, thinkingLevel,
     retryInfo, contextUsage, forkingEntryId,
     isCompacting, compactError, displayModel: displayModelValue, sessionStats,
-    agentPhase,
+    agentPhase, showScrollToBottom,
     isNew,
     messagesEndRef, scrollContainerRef,
-    lastUserMsgRef,
     handleSend, handleAbort, handleFork, handleNavigate, handleModelChange,
     handleCompact, handleSteer, handleFollowUp, handleAbortCompaction,
-    handleToolPresetChange, handleThinkingLevelChange, handleAgentEventRef,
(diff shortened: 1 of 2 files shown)


agegr:
Thanks for tackling #59.

I’m going to hold off on merging this PR as-is. The fix solves the jump-to-bottom problem, but it also removes the existing streaming-time bottom spacer and the behavior that anchors the latest user message near the top while the agent is responding. I’d like to keep that interaction model.

I’ll address this in a follow-up version with a smaller change: preserve the current streaming layout, but if the user has scrolled up to read earlier content, the chat should not jump to the latest message when the response finishes.

So I think the fix should be scoped to the completion-time auto-scroll, rather than replacing the existing streaming scroll behavior. Thanks again for the thoughtful work.

PR #75, opened 2026-06-11
fix: add rehype-raw plugin to render raw HTML tables in markdown content

fallleave001:
## Problem

Markdown content containing raw HTML `<table>` elements is not rendered
as tables in two places:

1. **Chat messages** (`MarkdownBody.tsx`): HTML tags are silently stripped
   by react-markdown because `rehype-raw` is not included.
2. **File preview** (`FileViewer.tsx`): Same issue in the markdown file
   preview (Preview mode).

## Fix

Add `rehype-raw` to the `rehypePlugins` array in both components:

- `components/MarkdownBody.tsx` — used for chat message rendering
- `components/FileViewer.tsx` — used for markdown file preview

Both already had `remarkGfm` / `rehype-katex` but were missing
`rehype-raw` for raw HTML passthrough.

## Testing

Confirmed that HTML `<table>` in `.md` files renders correctly in both
chat messages and file preview mode.

### DIFF
--- a/components/FileViewer.tsx
+++ b/components/FileViewer.tsx
@@ -6,6 +6,7 @@ import { vs } from "react-syntax-highlighter/dist/cjs/styles/prism";
 import { vscDarkPlus } from "react-syntax-highlighter/dist/cjs/styles/prism";
 import ReactMarkdown from "react-markdown";
 import remarkGfm from "remark-gfm";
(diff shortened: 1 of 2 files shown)


agegr:
已由 e974bc6 解决，并补了 rehype-sanitize 和依赖声明，所以不合并原 PR。

PR #76, opened 2026-06-11
fix: resolve Issue #56 - fix stack overflow on large session files

caiyongtian:
### Problem
Accessing sessions with large message history (e.g. 1068+ messages) caused HTTP 500 errors due to RangeError: Maximum call stack size exceeded during JSON serialization.

### Root Cause
- Massive session objects from buildSessionContext could not be serialized by NextResponse.json
- No error handling for corrupted session files
- No fallback scanning when SessionManager fails
- Deep agent state objects caused additional serialization issues

#### 1. lib/session-reader.ts
- Added 5-second TTL cache for listAllSessions() to avoid repeated expensive scans
- Implemented safeManualScan() - async fallback scanner using fs/promises
  - Reads only session headers (not full message history)
  - Prevents symlink loops with visitedDirs tracking
  - Non-blocking: runs asynchronously to not block the event loop
- Added sessions cache integration to resolveSessionPath()

#### 2. app/api/sessions/[id]/route.ts
- Granular try-catch blocks for: SessionManager.open, getEntries, getTree, getLeafId, getHeader, buildSessionContext, getSessionName, getRpcSession
- Message truncation: sessions with >500 messages are truncated to last 500
- Serialization safety net: pre-checks JSON.stringify, falls back to minimal response (empty tree, last 50 messages) if serialization fails
- Simplified agentState extraction: only includes isStreaming, isCompacting, contextUsage, systemPrompt, thinkingLevel (primitive-safe)
- Better error responses with details and stack traces

#### 3. next.config.ts
- Added '127.0.0.1' and 'localhost' to allowedDevOrigins to fix HMR WebSocket errors

### Testing
- Verified GET /api/sessions/019eb1da-0e62-775b-a744-520fb1da496d (1068 messages) now returns 200 with truncated data instead of 500 error
- Verified GET /api/sessions works correctly with caching
- Verified GET /api/sessions/:id?includeState works with simplified state

### DIFF
--- a/app/api/sessions/[id]/route.ts
+++ b/app/api/sessions/[id]/route.ts
@@ -21,13 +21,61 @@ export async function GET(
       return NextResponse.json({ error: "Session not found" }, { status: 404 });
     }
 
-    const sm = SessionManager.open(filePath);
-    const entries = sm.getEntries() as never;
-    const tree = sm.getTree();
-    const leafId = sm.getLeafId();
-    const context = buildSessionContext(entries, leafId);
+    let sm: any;
+    try {
+      sm = SessionManager.open(filePath);
+    } catch (openErr) {
+      console.error("[pi-web] Failed to open session file:", filePath, openErr);
+      return NextResponse.json({ error: "Failed to open session file. It may be corrupted." }, { status: 500 });
+    }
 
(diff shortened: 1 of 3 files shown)


agegr:
#85

PR #90, opened 2026-06-25
feat: show all recent projects in dropdown

fallleave001:
## Summary

Show all recently active project directories in the project dropdown instead of limiting the list to 5 items.

The dropdown now uses a bounded scroll container so long project lists remain usable without expanding indefinitely.

## Changes

- Remove the `.slice(0, 5)` limit in `getRecentCwds()`
- Make the dropdown vertically scrollable with `maxHeight: 320`
- Hide horizontal overflow to avoid long paths causing horizontal scroll

## Testing

- `npx tsc --noEmit`
- Verified the diff only touches `components/SessionSidebar.tsx`

### DIFF
--- a/components/SessionSidebar.tsx
+++ b/components/SessionSidebar.tsx
@@ -33,7 +33,7 @@ function formatRelativeTime(dateStr: string): string {
   return date.toLocaleDateString();
 }
 
-/** Return the 5 most recently active cwds across all sessions */
+/** Return all recently active cwds across all sessions, most recent first */
 function getRecentCwds(sessions: SessionInfo[]): string[] {
   const latestByCwd = new Map<string, string>(); // cwd -> most recent modified
   for (const s of sessions) {
@@ -45,7 +45,6 @@ function getRecentCwds(sessions: SessionInfo[]): string[] {
   }
   return [...latestByCwd.entries()]
     .sort((a, b) => b[1].localeCompare(a[1]))
-    .slice(0, 5)
     .map(([cwd]) => cwd);
 }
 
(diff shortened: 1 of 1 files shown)


agegr:
Thanks for the fix. Removing the 5-project limit is the right direction, but I don’t think the current scrollable dropdown style feels good enough yet. I’ll look for a cleaner UX before merging this.

ruggsea:
Dug into the harness-level half of this against real recorded data rather than reasoning about it.

**The display fix is confirmed correct against a real failover.** Found one on disk — `~/.pi/agent/sessions/--Users-ruggsea--/2026-08-05T11-59-18-721Z_019fd1ca-…jsonl`, rows 770-773:

| row | role | stopReason | api |
|---|---|---|---|
| 770 | user | — | — |
| 771 | assistant | `error` (content `[]`) | openai-completions |
| 772 | user (**identical text**) | — | — |
| 773 | assistant | toolUse | anthropic-**messages** |

The api flip on row 773 is the failover itself. `collapseFailoverRetries`'s structural rule (identical text, every assistant turn since the previous user message errored) matches this exactly.

One real gap found and closed in 8a66022: every existing test used **plain-string** content, but pi persists user content as an **array of text blocks** — the shape a reloaded session replays from disk. The array path was working, but nothing guarded it; all four string cases would still pass if it broke. Added the verbatim recorded shape as a fixture.

**The remaining transcript-level duplication cannot be fixed inside polypi.** `lib/rpc-manager.ts:465` re-runs the failed turn via `inner.prompt()` inside the chain loop, and each call appends a new user message. `AgentSession`'s public surface is `prompt / steer / followUp / abort / setModel / cycleModel / clearQueue` — there is no retry-without-recording primitive, and pi's session log is append-only, so the row cannot be retracted after the fact. The `auto_retry_start`/`auto_retry_end` events in `agent-session.d.ts` show pi *does* have an internal retry that doesn't re-record, but it is not drivable from outside and doesn't cover a model switch.

Worth noting this is slightly more than cosmetic: the duplicate stays in the context window permanently, so every later turn re-sends it.

So: **the fix belongs upstream in pi** — either a `retryLastTurn(model?)` public method, or letting the existing auto-retry path accept a model switch. Leaving this open as an upstream ask; nothing further to do on our side beyond the display collapse, which is verified.

PR #120, opened 2026-07-04
feat: fold assistant process details

zhudatou630:
## Summary
- Collapse completed intermediate assistant process messages under a Process details disclosure.
- Keep the final answer visible and keep the currently-running tail expanded.
- Hide empty completed thinking blocks and use English UI labels to match the existing app style.

## Why
Long tool-heavy conversations become hard to scan when every intermediate thinking/tool block stays fully expanded. Folding completed process details keeps the answer readable without losing access to the underlying work.

## Validation
- node --test lib/message-display.test.mjs
- npm run lint


zhudatou630:
Closing because upstream/main already contains an updated version of this feature, so this PR is now redundant.

PR #134, opened 2026-07-11
fix: normalize tool call blocks in API-loaded messages

fallleave001:
## Problem

The API returns tool call blocks with `id`/`name`/`arguments` fields, but UI components expect `toolCallId`/`toolName`/`input`. This causes `isEditToolName(undefined)` to crash when expanding Process Details (the function calls `.toLowerCase()` on `undefined`).

## Root Cause

`normalizeToolCalls` was only applied to streaming/completed messages, but not to messages loaded from the API via `setMessages(d.context.messages)`.

## Fix

Added `normalizeMessages()` helper that applies `normalizeToolCalls` to all messages loaded from the API (initial load and context reload).

## Testing

- TypeScript compilation passes
- Process Details expand/collapse works without crash
- Tool call blocks render correctly with normalized fields

### DIFF
--- a/hooks/useAgentSession.ts
+++ b/hooks/useAgentSession.ts
@@ -11,6 +11,13 @@ import type {
 } from "@/lib/types";
 import { normalizeToolCalls } from "@/lib/normalize";
 import { sendAgentCommand } from "@/lib/agent-client";
+
+/** Normalize tool call blocks in messages loaded from the API.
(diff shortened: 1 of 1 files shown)


agegr:
The client-side normalization is safe, but the stated root cause seems incomplete. Both session APIs use buildSessionContext(), which already calls normalizeToolCalls() in entryToUiMessage() before returning messages to the client.

PR #135, opened 2026-07-11
fix: disable single-dollar math rendering

fallleave001:
## Problem

`remarkMath` defaults to interpreting single dollar signs (`$`) as math delimiters, which breaks text containing currency values like `$10`. These get rendered as inline math instead of plain text.

## Fix

Set `singleDollarTextMath: false` so only double-dollar (`$$`) triggers math rendering, while single dollar signs are treated as literal text.


agegr:
`$x$` 这类行内公式是经常出现的。应该还是保留支持

PR #136, opened 2026-07-11
feat: optimize long session loading with trimmed context + paginated full history

fallleave001:
## Problem

Long sessions with compaction were slow to load because `buildSessionContext` iterated all entries and rendered every message, including compacted history.

## Solution

1. Use pi-coding-agent's trimmed context (`piCtx.messages`) as default
2. Add `buildFullHistory` API for paginated access to complete history
3. Add full history toolbar UI with navigation controls

### Backend (`lib/session-reader.ts`)
- `buildSessionContext`: use `piCtx.messages` directly (already trimmed)
- `buildFullHistory`: paginated full history with offset/limit
- `entryIndex`: lightweight metadata for all entries
- `MESSAGE_PRODUCING_TYPES`: filter for UI-relevant entries

### API (`app/api/sessions/[id]/history/route.ts`)
- New endpoint: `GET /api/sessions/[id]/history?offset=0&limit=200`
- In-memory cache with 5-min TTL and LRU eviction

### Client (`hooks/useAgentSession.ts`)
- `loadFullHistory`: paginated history loading
- `reloadTrimmedContext`: reload trimmed context
- `normalizeMessages`: fix tool call field names

### UI (`components/ChatWindow.tsx`)
- Full history toolbar with pagination controls
- Floating button for compaction sessions
- Request ordering to prevent stale responses

### Types (`lib/types.ts`)
- `EntryMeta`: lightweight entry metadata
- `entryIndex` in `SessionContext`

## Performance

| Metric | Value |
|--------|-------|
| Trimmed context | ~458 messages |
| Full history | ~11293 messages |
| Reduction ratio | ~24.7x |

## Verification

- ✅ TypeScript compilation passed
- ✅ Build successful
- ✅ EntryIds alignment verified
- ✅ Pagination working correctly


agegr (changes requested):
Thanks for working on long-session performance. Using the SDK's trimmed context for the default view and loading full history on demand is the right direction, but there are several correctness issues that should be addressed before merging.

### Blocking: `messages` and `entryIds` become misaligned after compaction

`buildSessionContext()` derives `entryIds` by taking the last N displayable entries:

```ts
convertibleEntries.slice(trimCount)
```

This does not match the SDK's compaction-aware ordering. The SDK returns:

1. The latest compaction summary
2. Entries starting from `firstKeptEntryId`
3. Entries after the compaction entry

For example, the returned values can be:

```text
messages: [summary, u2, a2, u3]
entryIds: [u2,      a2, cmp, u3]
```

The correct IDs are `[cmp, u2, a2, u3]`.

Since `entryIds` are used by Fork and `navigate_tree`, this can fork or navigate from the wrong entry. The mapping should be built from the SDK's actual context entries, for example using `buildContextEntries()` and `sessionEntryToContextMessages()`, rather than inferred from message counts.

### Full-history cache is not invalidated

The history endpoint caches all session entries for five minutes, but `invalidateSessionEntriesCache()` is never called. After full history has been opened once, continuing the conversation and navigating history can return an outdated total, leaf, and message list until the TTL expires.

The cache should either be invalidated on every session write or keyed/validated using the session file's mtime and size.

### Pagination still processes the complete history

`buildFullHistory()` converts every displayable message before applying `slice()`. Consequently, every page request still performs O(total history) processing and allocates the complete message array.

The implementation should build a lightweight path/index first, slice the requested entries, and only then convert those entries into UI messages.

The newly returned `entryIndex` is also currently unused by the UI and can make the default response substantially larger.

### Markdown sanitization is too broad

`stripInternalTags()` runs on every Markdown message and removes complete `<docId>`, `<read-files>`, and similar sections, including legitimate XML examples and code blocks.

This cleanup should be scoped to compaction summaries and should reuse the existing compaction-summary parser instead of modifying all rendered Markdown.

### Verification

`tsc --noEmit` passes, but `npm run lint` currently fails with four errors and two warnings. This change also needs focused tests covering:

- `messages`/`entryIds` alignment after compaction
- Retained entries before the compaction entry
- Multiple compactions
- Cache invalidation after appending messages
- Pagination boundaries and branch-specific leaves

The overall design is promising, but the entry-ID mismatch is a correctness blocker and should be fixed before merge.

agegr:
我先关了，先会按 #133 的建议优化一下长消息的性能

PR #140, opened 2026-07-12
Add optional password-only gate for public deployments

RrOrange:
Hi, thanks for building and maintaining pi-web.

I ran into a case where I wanted to expose pi-web publicly, but only enable a password gate when explicitly configured.

This PR is a small attempt to add an optional password-only login flow:

### What this changes
- Only enable auth when `PI_WEB_PASS` is set
- If `PI_WEB_PASS` is unset, pi-web behaves exactly as before
- When enabled, unauthenticated requests are redirected to `/login`
- The login UI asks for a password only (no username)
- After successful login, the user is redirected back to the original destination

### Implementation notes
- `middleware.ts` checks whether `PI_WEB_PASS` exists
- `/api/unlock` validates the password and sets an `HttpOnly` cookie
- `/login` provides a small password UI styled to fit the existing app theme

### Validation
I only did lightweight local validation here:
- `npm run lint`
- `npx tsc --noEmit`
- verified that `/` stays open when `PI_WEB_PASS` is unset
- verified that `/` redirects to `/login` when `PI_WEB_PASS` is set
- verified that correct/incorrect password handling works via `/api/unlock`

### Screenshot
Password login screen:

If this approach is not a fit for the project, no worries at all — I’d still be happy to adjust the implementation to better match your preferred direction.


agegr:
pi 权限过大，不支持直接公网暴露，建议使用虚拟组网、SSH 端口转发等方式。

PR #225, opened 2026-07-23
fix/223-minimap-search

Hamzaa6296:
Fix minimap hover flicker, differentiate message colors, add message search

- Add grace period on minimap mouse leave so tooltips don't disappear before the cursor can reach them
- Use distinct blue (user) / orange (assistant) colors for minimap dots instead of blue/gray
- Add MessageSearch component: search bar to find and jump to a specific user or assistant message in the chat

### DIFF
--- a/components/ChatMinimap.tsx
+++ b/components/ChatMinimap.tsx
@@ -1,6 +1,13 @@
 "use client";
 
-import { useEffect, useRef, useState, useCallback, useMemo, RefObject } from "react";
+import {
+  useEffect,
+  useRef,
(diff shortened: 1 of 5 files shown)


agegr:
Thanks for the contribution. I’m going to reject this PR for now.

It combines an unrelated message search feature with the minimap change, and most of the diff is formatting-only. More importantly, the minimap preview should show the selected node’s content, not depend on hover behavior.

Please split the search into a separate PR and rework the minimap interaction around node selection.

PR #229, opened 2026-07-24
feat: review agent changes by file or hunk

opsCar:
## Summary

- Capture Git working-tree changes made by each accepted agent prompt by snapshotting the selected repository scope before and after the run.
- Open a prompt-scoped Review panel when changes are detected, with file and hunk navigation plus Accept, Reject, Accept all, and Reject all actions.
- Apply review decisions transactionally with revision checks, file fingerprints, rollback, size limits, and safe file-level fallbacks for changes that cannot be reviewed by hunk.
- Bind review state to prompt run IDs so concurrent prompts and stale SSE events cannot attach changes to the wrong run.
- Re-authorize canonical allowed roots on every API operation and reject symlink escapes.

## Validation

- `npm run lint`
- `node --test lib/*.test.mjs components/*.test.mjs` — 117 tests passed
- `git diff --check origin/main...HEAD`

## Notes

- Code review requires a Git repository with an initial commit.
- New, deleted, renamed, binary, mode-changing, and oversized files fall back to whole-file review when safe.
- `next build` was intentionally not run, following the repository development guidance.

### DIFF
--- a/app/api/git/review/route.ts
+++ b/app/api/git/review/route.ts
@@ -1,0 +1,66 @@
+import { NextRequest, NextResponse } from "next/server";
+import { getAllowedFileRoots, isWindowsAbsolutePath } from "@/lib/file-access";
+import { activateGitReview, cancelGitReview, decideGitReview, finishGitReview, getGitReview, getGitReviewCwd, startGitReview } from "@/lib/git-review";
+import { resolveAllowedDirectory } from "@/lib/path-security";
+import type { GitReviewDecision } from "@/lib/git-types";
+
+type ReviewRequest =
+  | { action: "start"; cwd: string; runId: number }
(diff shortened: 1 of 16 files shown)


agegr:
Thanks for the thoughtful implementation and the attention to concurrency and safety. I don’t think this is ready to merge in its current form.

The feature is valuable, but it introduces a substantial Git review and patch-management subsystem into Pi Web. That feels heavier than the product’s current role as a web UI for Pi sessions, and it creates a large surface area around index state, conflicts, external edits, and rollback semantics. For example, rejecting a change currently restores the working tree but can leave agent-staged changes in the Git index.

I’d prefer a smaller first step, such as showing a read-only diff for changes made during a prompt. We can revisit accept/reject workflows once the desired Git semantics and long-term ownership are clearer.

PR #230, opened 2026-07-24
fix(agent/new): use randomUUID for new-session lock key to prevent session collision

### LINKED ISSUE #227: Bug: concurrent "New Session" requests in the same millisecond collide onto one shared agent session
## Summary

`POST /api/agent/new` derives its de-duplication lock key from `Date.now()` alone. Two new-session requests that arrive within the same millisecond receive the **same** `AgentSessionWrapper` and the **same** `sessionId`, so two independent chats are silently merged into one pi session.

## Root cause

`app/api/agent/new/route.ts:25`

```ts
// Use a one-time key so startRpcSession's lock doesn't conflict with real session ids
const tempKey = `__new__${Date.now()}`;
const { session, realSessionId } = await startRpcSession(tempKey, "", cwd, toolNames);
```

The comment says "one-time key", but `Date.now()` has millisecond resolution and is **not** unique under concurrency.

`startRpcSession` (`lib/rpc-manager.ts:1041-1042`) de-duplicates concurrent callers by that key via an in-flight lock:

```ts
const inflight = locks.get(sessionId);
if (inflight) return inflight;   // returns the SAME session promise
```

Creating a session (`createAgentSessionServices` → model/provider setup) takes tens to hundreds of ms, so the lock stays in-flight well beyond 1 ms. Any second `POST /api/agent/new` that computes the same `tempKey` gets the first request's in-flight promise and therefore the **same real session**.

## Steps to reproduce

1. Run `pi-web`.
2. Fire two new-session requests for the same `cwd` within the same millisecond. Either:
   - Open two browser tabs and start a new session in each at nearly the same time, or double-click "new session"; or
   - Reproduce deterministically with two near-simultaneous requests:
     ```bash
     curl -s -XPOST localhost:30141/api/agent/new \
       -H 'content-type: application/json' \
       -d '{"cwd":"/path/to/proj","type":"ensure_session"}' &
     curl -s -XPOST localhost:30141/api/agent/new \
       -H 'content-type: application/json' \
       -d '{"cwd":"/path/to/proj","type":"ensure_session"}' &
     wait
     ```
3. Compare the `sessionId` in both responses.

**Expected:** each request creates an isolated session with a distinct `sessionId`.

**Actual:** when the two requests land in the same millisecond, both responses carry the **same** `sessionId`. The two clients then drive one shared pi session — each client's first prompt is delivered into the other's conversation, and both read the same event stream.

## Impact

Session isolation failure / cross-conversation contamination. More likely under concurrent load or with multiple clients pointed at the same instance, and impossible for a client to detect or recover from.

## Suggested fix

Make the key actually unique (the code already imports `randomUUID` in `lib/rpc-manager.ts`):

```ts
import { randomUUID } from "crypto";
const tempKey = `__new__${randomUUID()}`;
```

A distinct key per request means the in-flight lock only ever coalesces retries of the *same* logical request, never two different ones. Happy to send a PR.

Cyriellewu:
## Summary

`POST /api/agent/new` derived its de-duplication lock key from `Date.now()` alone, which has only millisecond resolution. `startRpcSession()` (`lib/rpc-manager.ts`) coalesces concurrent callers by that key via an in-flight lock, so two new-session requests arriving in the **same millisecond** received the same `AgentSessionWrapper` and the same `sessionId` — silently merging two independent chats into one pi session.

Fixes #227

## Root cause

`app/api/agent/new/route.ts`:
```ts
const tempKey = `__new__`;
```
Session creation takes tens to hundreds of ms, so the in-flight lock stays live well beyond 1 ms. Any second `POST /api/agent/new` computing the same `tempKey` gets the first request's in-flight promise, and therefore the same real session — a cross-conversation isolation failure the client cannot detect or recover from.

## Fix

Use `randomUUID()` for the key (already imported and used the same way throughout `lib/rpc-manager.ts`):
```ts
const tempKey = `__new__`;
```
A distinct key per request means the in-flight lock only ever coalesces retries of the *same* logical request, never two different ones. `__new__`/`tempKey` has exactly one call site in the repo, and nothing depends on the key being time-ordered or parseable, so there is no downstream impact.

## Testing

The repo has no test harness, so this was verified by inspection and the deterministic reproduction from the issue (two near-simultaneous `POST /api/agent/new` now return distinct `sessionId`s). The change is a drop-in of an existing, already-used pattern.

### DIFF
--- a/app/api/agent/new/route.ts
+++ b/app/api/agent/new/route.ts
@@ -1,5 +1,6 @@
 import { NextResponse } from "next/server";
 import { existsSync } from "fs";
+import { randomUUID } from "crypto";
 import { allowFileRoot } from "@/lib/file-access";
 import { invalidateSessionListCache } from "@/lib/session-reader";
 import { startRpcSession } from "@/lib/rpc-manager";
@@ -19,10 +20,14 @@ export async function POST(req: Request) {
       return NextResponse.json({ error: `Directory does not exist: ${cwd}` }, { status: 400 });
     }
 
-    // Use a one-time key so startRpcSession's lock doesn't conflict with real session ids
     const { provider, modelId, toolNames, thinkingLevel, ...promptCommand } = command as { provider?: string; modelId?: string; toolNames?: string[]; thinkingLevel?: string; [key: string]: unknown };
 
-    const tempKey = `__new__${Date.now()}`;
+    // Use a unique one-time key so startRpcSession's in-flight lock never
+    // coalesces two independent new-session requests. Date.now() has only
+    // millisecond resolution, so concurrent calls in the same millisecond
+    // would otherwise share one lock key and be merged into a single pi
+    // session (see issue #227). randomUUID() guarantees per-request isolation.
+    const tempKey = `__new__${randomUUID()}`;
     const { session, realSessionId } = await startRpcSession(tempKey, "", cwd, toolNames);
 
     // Keep the files-route allowed-roots cache (see app/api/files/[...path]/route.ts)


agegr:
already merged in #228

PR #238, opened 2026-07-24
feat: add recent cwd path selection

kerwin2046:
Problem: Users who frequently switch between working directories had to re-enter custom paths or scroll through all session-derived projects to find their desired directory. There was no persistence of manually entered or recently used paths — they were lost after navigating away.
Changes:
1. New API endpoint app/api/cwd/recent/route.ts
- GET /api/cwd/recent — returns the list of recent working directories from ~/.pi/agent/recent-cwds.json
- POST /api/cwd/recent — accepts { cwd: string }, prepends it to the recent list (deduplicated), caps at 10 entries, and persists to the JSON file
2. components/SessionSidebar.tsx
- On mount, fetches the recent paths list and stores it in local state
- Whenever the user selects a CWD (via project list, custom path, default directory, or worktree switch), a debounced effect records it as a recent path via POST /api/cwd/recent
- The CWD dropdown now includes a "Recent" section between the project list and the "Default directory"/"Custom path" actions
- Recent paths that already appear in the current project list are filtered out to avoid duplication
- Up to 5 recent paths are shown, each with a folder icon and left-ellipsized path label
- The "Recent" section is hidden when the project filter is active to avoid confusion

### DIFF
--- a/app/api/cwd/recent/route.ts
+++ b/app/api/cwd/recent/route.ts
@@ -1,0 +1,64 @@
+import { NextResponse } from "next/server";
+import { readFileSync, writeFileSync, existsSync, mkdirSync } from "fs";
+import { join, resolve } from "path";
+import { homedir } from "os";
+import { getAgentDir } from "@earendil-works/pi-coding-agent";
+
+const MAX_RECENT = 10;
+
+function getRecentPath(): string {
+  const agentDir = getAgentDir();
+  return join(agentDir, "recent-cwds.json");
+}
+
+function readRecent(): string[] {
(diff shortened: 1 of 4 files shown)


kerwin2046:
image

agegr:
Thanks for the contribution. The current CWD selector already lists session-backed projects ordered by recent activity, with filtering for larger lists.

We’d like to keep the selector scoped to directories that have sessions, so we don’t need a separate persisted “Recent paths” list at this time. We’ll close this PR for now. Thanks for understanding.

PR #239, opened 2026-07-24
feat(markdown): support LaTeX bracket delimiters

GuangmingChan:
## Problem

Pi Web already renders KaTeX for `$...$` and `$$...$$`, but model
responses commonly use `\(...\)` and `\[...\]`. These delimiters were
displayed as plain text.

## Changes

- Normalize `\(...\)` to inline math syntax.
- Normalize `\[...\]` to display math syntax.
- Preserve delimiters inside inline and fenced code blocks.
- Add rendering and regression tests.

## Validation

- `node --test components/MarkdownBody.test.mjs`
- `npx tsc --noEmit`
- `npm run build`

## 简言之，我修改了数学公式渲染功能，请审核。如果修改合适，请把它合并到官方版本。

### DIFF
--- a/components/MarkdownBody.test.mjs
+++ b/components/MarkdownBody.test.mjs
@@ -35,3 +35,26 @@ test("keeps local file markdown links in the app", () => {
   assert.match(html, /<a href="components\/MarkdownBody\.tsx">file<\/a>/);
   assert.doesNotMatch(html, /target=|rel=|\snode=/);
 });
+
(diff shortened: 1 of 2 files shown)


agegr:
Support has now been implemented on main in commit 65d59dc using the shared Markdown normalizer for both chat and file preview. The implementation only converts complete delimiter pairs and preserves code, escaped literals, raw HTML, and link destinations.
Thanks again for the contribution.

PR #243, opened 2026-07-24
feat: enhance model validation and error handling

kerwin2046:
Bug: models.json schema validation is all-or-nothing — one invalid model entry silently drops ALL custom providers
Reproduction:
1. Defines multiple custom providers in ~/.pi/agent/models.json, e.g., new-provider-xx (3 models),new-provider-2xx  (3 models), new-provider-3xx (1 model)
2. One of the model entries has an empty id field: { "id": "" }
3. The entire model dropdown in the UI shows zero custom models — only built-in providers appear
4. No error or warning is shown anywhere
Root cause:
ModelConfig.load() in pi-coding-agent uses Typebox schema validation on the entire models.json file. If ANY single field across ANY provider violates the schema (e.g., Type.String({ minLength: 1 }) rejects ""), the whole file is discarded and ModelConfig returns an empty providers Map. ModelRuntime.rebuildProviders() then has no custom providers to register, so modelRuntime.getAvailable() returns only builtins — silently.
The relevant code path:
models.json → ModelConfig.load() → validateModelsConfig.Check(parsed) → false → new ModelConfig(new Map(), error)
                                                                          ↘ true  → new ModelConfig(parsedProviders)
Why it's insidious:
- The error is stored in ModelConfig.error and exposed via ModelRuntime.getError(), but neither /api/models nor the UI ever read it
- Users see "model dropdown is empty for my custom providers" with no clue why
- A typo in one model definition poisons every other correctly-configured provider in the same file
- The failure is silent and undebuggable without reading server logs
Fix (pi-web side, defense-in-depth):
- PUT /api/models-config: Inline validation before writing — checks each model's id is a non-empty string, provider structure is valid. Returns HTTP 400 with a clear message like Provider "new-provider-xx", model #3: "id" must be a non-empty string. The ModelsConfig modal shows this inline in its footer.
- GET /api/models: Checks modelRuntime.getError() specifically for schema/parse errors (Invalid models.json schema: / Failed to parse models.json:). Only these block model loading; availability/composition errors still allow partial results. Returns the error as configError in the response for upstream consumers.

image


agegr:
Thanks for investigating this issue and identifying the upstream validation behavior. We’re going with a smaller fix in pi-web: display `ModelRuntime.getError()` while keeping available built-in models visible.

This PR also includes unrelated changes and duplicates existing model reload work, so we won’t merge it. Thanks again for the detailed report and contribution.

PR #248, opened 2026-07-25
feat: add password auth and mobile PWA support

sleepinginsummer:
## Summary

- add first-run password setup, scrypt password verification, signed 30-day sessions, and route protection
- support both enhanced JSON login and native form POST fallback with safe same-origin redirects
- add an installable PWA manifest, opaque wordmark icons, offline fallback, and controlled service worker updates
- show an update prompt instead of refreshing an active workspace automatically
- render the initial mobile layout from the request user agent to avoid hydration layout shifts
- show the Get Started hint once per browser and document the new authentication/PWA structure

## Security

- stores only a salted scrypt hash and random session secret in a mode `0600` config file
- uses `HttpOnly`, `Secure` (behind HTTPS), `SameSite=Lax` cookies
- rejects expired and tampered session tokens with timing-safe signature comparison
- keeps API routes protected while exposing only login and required PWA assets

## Testing

- `npm run lint`
- `node --check public/sw.js`
- `node --test lib/auth.test.mjs`


agegr:
Thanks for the contribution. Since main has changed significantly and this branch now conflicts with the latest security-related changes, I’ll close this PR for now. I’ll follow up by implementing the relevant parts directly on the latest main, using this PR as a reference.

PR #250, opened 2026-07-25
feat: add an optional live bot bridge

coinwh:
Hi, and thank you for maintaining Pi Web.

This PR proposes an optional bot bridge for users who would like to follow and control a Pi session from Telegram. The feature is disabled by default and can be configured from a new **Bridge** button in the top action bar.

### What it adds

- Server-side bot-token storage with only a masked hint exposed to the browser
- Allow-listed chat IDs and a configurable working directory
- One persistent Pi session per chat
- Live message updates for the model name, Thinking output, tool calls and arguments, tool duration, token/cache usage, cost, and streamed response text
- Throttled message editing to avoid Telegram API rate limits
- An optional guard that rejects new messages while the model is responding
- `/new`, `/status`, `/model`, `/stats`, `/stop`, and `/help` commands
- Responsive Bridge settings UI integrated with the existing Pi Web styling
- English and Chinese README documentation

The implementation keeps the integration opt-in and local: enabling it requires an explicit token, at least one allowed chat ID, and a working directory.

I would be grateful if you would consider including this feature. I am happy to adjust the UI, naming, behavior, or implementation details to better match the project's direction. Thank you for your time and review.

### Testing

Not run in this iteration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ npx @agegr/pi-web@latest
 
 ## Features
 
+- **Telegram bridge** — connect an allow-listed Telegram user or group to Pi. Configure the bot token, chat IDs, and working directory from the Telegram panel in the sidebar; each chat keeps an independent Pi session and supports `/new`, `/status`, and `/help`.
+
 - **Pick work back up**: browse previous pi conversations by project without digging through terminal history or session paths.
 - **Try different directions safely**: continue from an earlier message or fork a session into a separate route.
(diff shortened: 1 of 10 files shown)


coinwh:
image
屏幕截图 2026-07-24 201105
image

agegr:
Thank you for the thoughtful contribution. The feature is useful, but we don’t think a Telegram-specific bridge belongs in Pi Web’s core scope at this time. It would be better maintained as a separate extension or companion package.

PR #258, opened 2026-07-25
feat: pinned cross-project recent sessions in sidebar

rollysys:
## Problem

Sessions are grouped by project in the sidebar. Jumping between sessions that live in **different projects** takes three steps every time: open the project dropdown → find the project → find the session. For users who hop between projects constantly (agent workflows make this the norm), switching sessions is the most repeated navigation action and the current grouping makes it tedious.

## Solution

A collapsible **RECENT** section pinned at the top of the session list:

- The **8 most recently modified sessions across all projects**, sorted by `modified` desc
- Each row shows a **project label** (folder icon + basename of `projectRoot ?? cwd`) so same-titled sessions in different projects stay distinguishable
- **One click** switches the project and opens the session — it reuses the existing `handleSelectSessionFromList` cross-project handler, so selection highlight, running/unread indicators, rename and delete all behave exactly as in the per-project tree (same `SessionItem` component)
- The section is collapsible (default expanded); the per-project tree below is completely unchanged

`SessionItem` gains one optional `projectLabel` prop rendered in the meta line; no other component touched. Single-file change: `components/SessionSidebar.tsx` (+84/-1).

## Testing

- [x] `next build` passes
- [x] Sidebar renders RECENT with sessions from multiple projects in recency order
- [x] Clicking a recent entry from another project switches the project dropdown, loads the session, and updates the URL
- [x] Rename/delete from within the RECENT section work and refresh both lists
- [x] Collapse toggle persists for the session

### DIFF
--- a/components/SessionSidebar.tsx
+++ b/components/SessionSidebar.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useEffect, useLayoutEffect, useState, useCallback, useRef, type CSSProperties, type ReactNode } from "react";
+import { useEffect, useLayoutEffect, useMemo, useState, useCallback, useRef, type CSSProperties, type ReactNode } from "react";
 import type { SessionInfo } from "@/lib/types";
 import { FileExplorer, type FileExplorerHandle } from "./FileExplorer";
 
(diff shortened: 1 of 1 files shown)


agegr:
Thanks for the contribution and for exploring this workflow.

I tested the Recent sessions design locally. While it makes cross-project switching faster, it also adds another navigation concept alongside the existing project and worktree selectors. In practice, the extra mode/section increases sidebar complexity and duplicates sessions that are already available in the project tree.

After considering a few alternative layouts, I think it is better to keep the current sidebar model focused and not add the Recent sessions feature for now. I’m going to close this PR, but I appreciate the thoughtful implementation and the work that went into it.

PR #276, opened 2026-07-26
fix: gate untrusted project extensions behind project trust (#236)

### LINKED ISSUE #236: Project-local `.pi` extensions execute local commands during pi-web session startup without project trust confirmation
## Summary

pi-web loads and executes project-local Pi extensions from a repository during normal session startup without showing a project trust confirmation.

A repository can include a `.pi/extensions/*.js` file. When a user creates a new pi-web session for that repository, pi-web calls `createAgentSessionServices({ cwd, agentDir })` without passing the Pi SDK project-trust resolver. As a result, the project extension is loaded and its `session_start` handler can execute local commands as the user running pi-web.

This is not a prompt-injection issue. The PoC does not require model output, tool approval, or the model choosing to run a command.

## Affected version

Tested with:

- Repository: `agegr/pi-web`
- pi-web version: `@agegr/pi-web` `0.7.12`
- Source revision inspected: `036cb7c0c20f0db8304a047462b7ea4cd9986a2c`
- `@earendil-works/pi-coding-agent`: locked to `0.80.7` in `package-lock.json`
- Environment: Linux x86_64, Node/npm local checkout

## Impact

An attacker-controlled repository can execute local JavaScript and local commands when a victim creates or restores a pi-web agent session for that repository.

The execution happens in the local pi-web/Node process context, with the permissions of the user running pi-web. The benign PoC below only writes a marker under `/tmp`, but the same primitive can run arbitrary commands available to that local user.

## Proof of concept

Use a disposable repository containing this project-local extension:

```js
// .pi/extensions/session-start-marker.js
export default (pi) => {
  pi.on("session_start", async () => {
    await pi.exec(process.execPath, [
      "-e",
      "require('node:fs').writeFileSync('/tmp/pi-web-project-extension-autoexec-marker','executed\\n')",
    ]);
  });
};
```

Start pi-web with an isolated Pi agent directory:

```bash
cd /path/to/pi-web
npm ci
PI_CODING_AGENT_DIR=/tmp/pi-web-agent ./node_modules/.bin/next dev -p 30142 -H 127.0.0.1
```

Trigger normal session creation for the PoC repository:

```bash
curl -sS -X POST http://127.0.0.1:30142/api/agent/new \
  -H 'Content-Type: application/json' \
  --data '{"cwd":"/path/to/poc-repo","type":"ensure_session","toolNames":[]}'
```

Observed result:

```bash
cat /tmp/pi-web-project-extension-autoexec-marker
# executed
```

I reproduced this with the safe fixture at:

```text
poc/pi-web-project-extension-autoexec/
```

The helper output was:

```text
SUCCESS: marker written at /tmp/pi-web-project-extension-autoexec-marker
executed
```

## Expected behavior

pi-web should detect project-local trust-requiring resources such as `.pi/extensions` and `.pi/settings.json` before loading them, and should either prompt for project trust or refuse to load them until trust is granted.

Repository-controlled extension code should not be imported, evaluated, or allowed to receive `session_start` before that trust decision.

## Root cause

In `lib/rpc-manager.ts`, `startRpcSession()` creates Pi SDK services with only:

```ts
const services = await createAgentSessionServices({ cwd, agentDir });
```

No `resourceLoaderReloadOptions.resolveProjectTrust` callback is provided.

The resulting flow is:

```text
POST /api/agent/new
-> startRpcSession(cwd)
-> createAgentSessionServices({ cwd, agentDir })
-> resourceLoader.reload(...)
-> project `.pi/extensions` resource is loaded
-> jiti.import(extensionPath)
-> extension factory runs
-> bindExtensions()
-> session_start handler runs
-> pi.exec(...)
```

pi-web also starts/restores sessions from saved session files by reading the session header `cwd` and calling `startRpcSession(id, filePath, cwd)`, so restored sessions should be protected by the same trust gate.

## Why this crosses a security boundary

Project extensions are executable code controlled by the repository. Creating a pi-web agent session for a repository is not the same as approving repository-provided extension code to run locally.

The Pi SDK appears to have a project-trust mechanism for resources such as `.pi/settings.json`, `.pi/extensions`, `.pi/skills`, `.pi/prompts`, `.pi/themes`, `.pi/SYSTEM.md`, `.pi/APPEND_SYSTEM.md`, and ancestor `.agents/skills`. The issue is that pi-web's web session creation/restoration path does not invoke that resolver before resource loading.

## Suggested remediation

1. Pass the SDK project-trust resolver into `createAgentSessionServices()` from pi-web's `startRpcSession()` path.
2. Require an explicit trust decision before loading any repository-controlled executable project resources, including `.pi/extensions` and extension entries from project `.pi/settings.json`.
3. Enforce the trust decision before module import, extension factory invocation, missing package installation, or `session_start` dispatch.
4. Bind the approval to the canonical workspace path and consider content-hash re-confirmation when executable extension declarations or referenced files change.

shani-singh1:
## Summary

Fixes #236.

Opening a repository in Pi Web executes that repository's project-local extensions. Building session services imports every enabled `.pi/extensions` module (running its factory), and the session then dispatches `session_start` handlers. Because Pi Web called `createAgentSessionServices({ cwd, agentDir })` **without** the SDK's project-trust resolver, merely creating a session — or even enumerating models — for an untrusted repo ran repository-controlled code locally, with no trust prompt.

This is not prompt-injection: it needs no model output and no tool approval. Per the reproduction in #236, a `.pi/extensions/*.js` `session_start` handler runs on `POST /api/agent/new`.

## Fix

New helper `lib/project-trust.ts` returns the SDK's `resourceLoaderReloadOptions` **only** when the project actually has trust-requiring resources (`hasTrustRequiringProjectResources`), and resolves the decision from the `ProjectTrustStore` that is shared with the `pi` CLI, defaulting to **untrusted**:

```ts
export function projectTrustReloadOptions(cwd, agentDir) {
  if (!cwd || !hasTrustRequiringProjectResources(cwd)) return undefined;
  const trustStore = new ProjectTrustStore(agentDir);
  return { resolveProjectTrust: async () => trustStore.get(cwd) === true };
}
```

The SDK's resource loader runs a pre-trust pass that loads only user/global extensions, calls `resolveProjectTrust`, and includes project-local extensions **only if it resolves `true`** (`core/resource-loader.js`). So denying trust keeps the repo's extensions from ever being imported or bound.

Wired into both code paths that create services from a project `cwd`:
- `lib/rpc-manager.ts` `startRpcSession` (session create/restore — the `session_start` path)
- `app/api/models/route.ts` (model enumeration still imports project extension factories)

### Behavior

- Project with no `.pi` trust-requiring resources → `undefined`, unchanged (fast) load path.
- Project already trusted in the `pi` CLI → `trustStore.get(cwd) === true` → extensions load as before (the store is shared).
- Untrusted project → extensions stay dormant; **no auto-execution.**

### Scope / follow-up

Pi Web has no in-app trust prompt yet, so a project is currently trusted by trusting it once in the `pi` CLI. A web trust-prompt UI (the SDK exposes `getProjectTrustOptions` / `ProjectTrustStore.set`) is a good follow-up; this PR is the minimal change that closes the code-execution hole while honoring existing trust decisions.

## Testing

- `tsc --noEmit` passes
- `eslint` passes
- `node --test lib/project-trust.test.mjs` passes — asserts the helper defaults to deny for untrusted projects, returns `undefined` for clean ones, and that both session-creation sites forward the gate (source-inspection style, matching `rpc-manager.test.mjs`, to avoid loading the full SDK runtime in unit tests).

Verification of the SDK trust mechanics was done by reading `@earendil-works/pi-coding-agent` (`core/resource-loader.js` pre-trust pass, `core/trust-manager.js`) and mirroring the resolver pattern the `pi` CLI itself uses in `main.js`.

### DIFF
--- a/app/api/models/route.ts
+++ b/app/api/models/route.ts
@@ -4,6 +4,7 @@ import { createAgentSessionServices, getAgentDir, type SettingsManager } from "@
 import { getSupportedThinkingLevels } from "@earendil-works/pi-ai";
 import { loadModelsWithCache, withModelRuntimeError, type ModelsData } from "@/lib/models-cache";
 import { getAllowedFileRoots, isExistingFilePathAllowed } from "@/lib/file-access";
+import { projectTrustReloadOptions } from "@/lib/project-trust";
 
 export const dynamic = "force-dynamic";
 
@@ -47,7 +48,15 @@ async function loadModels(cwd: string): Promise<ModelsData> {
   const thinkingLevelMaps: Record<string, Record<string, string | null>> = {};
 
   const agentDir = getAgentDir();
-  const services = await createAgentSessionServices({ cwd, agentDir });
+  // Gate untrusted project extensions: enumerating models still imports and
+  // runs a repository's .pi/extensions factories, so honor project trust here
+  // too (see lib/project-trust.ts, #236).
+  const trustReloadOptions = projectTrustReloadOptions(cwd, agentDir);
+  const services = await createAgentSessionServices({
+    cwd,
+    agentDir,
+    ...(trustReloadOptions ? { resourceLoaderReloadOptions: trustReloadOptions } : {}),
+  });
   const available = await services.modelRuntime.getAvailable();
   const modelError = services.modelRuntime.getError();
   const settings: SettingsManager = services.settingsManager;
--- a/lib/project-trust.test.mjs
+++ b/lib/project-trust.test.mjs
@@ -1,0 +1,31 @@
+import assert from "node:assert/strict";
+import { readFile } from "node:fs/promises";
+import test from "node:test";
+
+// This module gates repository-controlled code execution, so the wiring is
+// asserted by source inspection (like rpc-manager.test.mjs) rather than by
+// loading the full SDK runtime.
+
+test("projectTrustReloadOptions denies untrusted projects and skips clean ones", async () => {
+  const source = await readFile(new URL("./project-trust.ts", import.meta.url), "utf8");
+
+  // Detects trust-requiring project resources via the SDK helper.
+  assert.match(source, /hasTrustRequiringProjectResources\(cwd\)/);
+  // No such resources -> undefined, so ordinary projects keep the normal path.
+  assert.match(source, /return undefined/);
+  // Otherwise the decision comes from the shared CLI trust store, default deny.
+  assert.match(source, /new ProjectTrustStore\(agentDir\)/);
+  assert.match(source, /resolveProjectTrust:\s*async \(\) => trustStore\.get\(cwd\) === true/);
+});
+
+test("both session-creation sites pass the project-trust gate to the SDK", async () => {
+  for (const rel of ["./rpc-manager.ts", "../app/api/models/route.ts"]) {
+    const source = await readFile(new URL(rel, import.meta.url), "utf8");
+    assert.match(source, /projectTrustReloadOptions\(cwd, agentDir\)/, `${rel} should compute trust reload options`);
+    assert.match(
+      source,
+      /resourceLoaderReloadOptions: trustReloadOptions/,
+      `${rel} should forward the gate to createAgentSessionServices`,
+    );
+  }
+});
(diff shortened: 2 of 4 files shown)


agegr:
merged in #280

PR #278, opened 2026-07-27
fix: share running-sessions SSE across tabs via BroadcastChannel

dingxiaobo:
Each open tab opened its own /api/agent/running/events SSE. Browsers cap HTTP/1.1 same-origin connections at 6 and SSE holds them open, so ~3 tabs starved the pool and new agent/file-watch SSE could no longer connect.

The running list is a global set identical for every tab, so only one tab needs to hold the stream. Leader election via a localStorage lease (1.5s heartbeat, 4s stale) picks the leader; the leader relays frames over a BroadcastChannel and followers reuse them. Per-tab agent events SSE is per-session and untouched. BroadcastChannel-less browsers fall back to the original per-tab behavior.

Side effect: SessionSidebar drops the sseAuthoritativeRef in favor of an applyFallback callback that mirrors the same "stream is authoritative once it delivered a frame" guard.

Effect: N tabs hold 1 running SSE instead of N; tab budget under the 6-connection limit roughly doubles.

### DIFF
--- a/components/SessionSidebar.tsx
+++ b/components/SessionSidebar.tsx
@@ -4,6 +4,7 @@ import { useEffect, useLayoutEffect, useState, useCallback, useRef, type CSSProp
 import type { SessionInfo } from "@/lib/types";
 import { DirectoryPicker } from "./DirectoryPicker";
 import { FileExplorer, type FileExplorerHandle } from "./FileExplorer";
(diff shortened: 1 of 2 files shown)


dingxiaobo:
Not fixed, close.

PR #292, opened 2026-07-29
Localize full history exports

FlowersYu:
## What changed

- pass the active Pi Web locale to the full-history export endpoint
- localize exporter-owned standalone HTML UI with the existing message catalog
- keep English exports and encoded session content unchanged
- add tests against the real `pi-coding-agent` export templates

## Why

The Full history button uses Pi Web's React i18n layer, but the opened page is generated by `pi-coding-agent` outside the React tree, so its controls and summary labels remained English.

## Validation

- `npm run lint`
- `node --test --test-reporter=spec lib/*.test.mjs components/*.test.mjs` (169/169)
- real Chinese and English export requests returned 200 with identical encoded session payloads
- browser smoke test covered Chinese rendering, user filter interaction, English fallback, and console errors


agegr:
Thanks for the contribution. I don’t think we should merge this approach.
The standalone export HTML and its UI strings are owned by pi-coding-agent, so localization should be implemented upstream. Rewriting the generated HTML in Pi Web through exact string replacements tightly couples us to an internal upstream template: template changes can silently leave untranslated text or break the exported page. It also requires repeatedly scanning the complete HTML, including the Base64-encoded session payload, which causes substantial memory overhead for large sessions.
The preferred solution is for pi-coding-agent to expose a locale or message-catalog option when exporting. Pi Web can then pass its active locale to that API.
Thanks again for investigating and validating the localization workflow.

PR #294, opened 2026-07-29
Add retry and model settings actions for prompt errors

Atopos-jy:
## Summary

Improve the recovery flow when a prompt fails.

- Keep prompt error notices visible until dismissed.
- Add a **Retry** action for the failed text prompt.
- Add a **Check models** action that opens the model configuration dialog.
- Add English and Simplified Chinese labels for the new actions.

## Why

A failed model request currently shows a temporary notice that disappears after a few seconds. Users can be left with a message that appears unanswered and no direct way to retry or inspect model configuration.

## Validation

- `npm run lint`

### DIFF
--- a/components/AppShell.tsx
+++ b/components/AppShell.tsx
@@ -1400,6 +1400,7 @@ export function AppShell() {
               onSessionStatsPanelOpen={openSessionStatsPanel}
               onContextUsageChange={handleContextUsageChange}
               onOpenFile={handleOpenLinkedFile}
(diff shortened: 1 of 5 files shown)


agegr:
Thanks for working on this. Since this PR was opened, main has added persistent provider-error rendering in the message area via 556e2ec (`fix: surface provider and compaction errors`).

For the current scope, displaying the error in the conversation is sufficient; retry and model-settings actions are not needed. Because the error-display requirement is already covered on main, I don't think we need to merge this PR.

PR #299, opened 2026-07-29
feat(chat): file-preview links for bare paths and turn artifacts

thy950523:
## Why

When the agent produces a file (a generated report, an HTML page), there's usually no one-click way to preview it: the agent's summary tends to omit the absolute path, so even though the file exists there's nothing to click. This adds two small, complementary ways to open those files in the right-hand preview pane — and neither edits the reply text.

## What

1. **Bare-path links** — absolute file paths written in a reply become clickable in-app preview links.
2. **Turn-artifact chips** — one chip per file produced by the turn's successful `write`/`edit` tool calls, shown under the reply (even when the reply never names the file).
3. `.html` files open in rendered-preview mode (live page via a sandboxed iframe) instead of source.

## How

- **Bare-path links = a remark plugin.** It walks markdown text nodes, finds absolute paths with a previewable extension, skips code / inline-code / URLs, and wraps matches in link nodes. Clicking reuses the existing link handler → preview pane.
- **Chips = a pure extractor + one leaf component.** `extractTurnArtifacts(turnContent, toolResults, cwd)` keeps the turn's `write`/`edit` tool calls whose result succeeded (skips errored / not-yet-arrived, dedupes), resolves paths against `cwd`; `TurnArtifacts` renders one `<button>` per file → the same existing preview path.

## Review focus — where it hooks in, and why there

Each feature is placed at the **narrowest layer that already has the data it needs**, so the blast radius is minimal:

- Bare-path links → **markdown-AST layer only** (one line registers the plugin in `lib/markdown.ts`; everything else is a pure transform). A path in reply text is purely a *rendered-text* concern, so this never touches message data.
- Chips → **turn-orchestration layer** (`components/ChatWindow.tsx`) + a leaf component. This needs *cross-message* data: Pi records each tool call as its own message and the reply as a separate final message, so the extractor must aggregate the whole turn before it can see the tool calls.

**Ask:** please confirm you're comfortable with (a) bare-path links living entirely inside the remark pipeline, and (b) turn-artifact aggregation happening in `ChatWindow`. Everything below these layers — session reading / normalization, the per-tool-call message split, the backend file-access gate (`isFilePathAllowed`) — is untouched. No new attack surface: both features route through the existing trusted preview path.

## Changes & verification

23 files, +2323 / −40.
**New:** `lib/file-path-scan.ts`, `lib/remark-file-paths.ts`, `lib/turn-artifacts.ts`, `components/TurnArtifacts.tsx` (+ unit tests).
**Edited:** `lib/markdown.ts` (register plugin), `components/ChatWindow.tsx` (turn aggregation), `components/MessageView.tsx` (prop + render), `components/FileViewer.tsx` (HTML preview default).

Automated: `node --test` — lib 170/171, components 26/26; `tsc --noEmit` and `eslint` clean.

Manual (fresh session):

**1. Chip + HTML rendered preview** — agent created `demo.html`; the chip opens it in Preview mode (live page, not source).
1 — chip + HTML preview

**2. Bare path → link** — an absolute path written in the reply is clickable.
2 — bare path becomes a link

**3. Path in a code block stays plain** — only prose paths linkify, so code stays copy-pasteable.
3 — code-block path not linkified

🤖 Generated with [Claude Code](https://claude.com/claude-code)


agegr:
Thanks for the contribution. The successful `write`/`edit` artifact entries and opening HTML files in Preview mode are useful improvements. However, we will not merge this PR in its current form.

There are two main reasons:

1. Bare absolute paths should not be linkified automatically. Passive assistant text should not become a file-opening action. File links should remain explicit, or be derived from successful file-producing tool calls. The bare-path behavior does not match the current safety and interaction design.
2. The PR is much broader than the accepted functionality. It combines three separate behaviors with a path scanner, Markdown transformation, file-type/API refactoring, four plan/spec documents, and screenshots. This makes the change unnecessarily difficult to review and maintain.

A focused replacement PR would be welcome with only:

- deduplicated entries for files successfully handled by `write`/`edit` in the current turn, shown under the final response;
- HTML files opening in Preview mode by default;
- focused tests for those behaviors.

Please exclude bare-path linkification, the associated scanner/Markdown changes, generated plan/spec documents, screenshots, and unrelated refactoring.

For those reasons, this version is rejected, but the two scoped features above are welcome in a smaller replacement PR.

PR #300, opened 2026-07-29
fix: surface assistant error messages in the chat UI

adlternative:
## Problem

When the model request fails, pi stores an assistant message with `stopReason: "error"` and an `errorMessage`, but with empty `content`. The web UI rendered nothing for these messages:

- `AssistantMessageView` has an early return for empty-content messages (`MessageView.tsx`), so the whole message was dropped;
- the turn-grouping logic in `ChatWindow` only renders a final answer bubble when there are answer blocks, so an errored final message produced no bubble either;
- `useAgentSession` raised no notice on `message_end` for errored messages.

The result: on provider failures (e.g. `"All backends failed"`, context overflow, queue errors) the agent appears to silently stop / "disconnect", and the only way to find out what happened is to read the session `.jsonl` by hand. The TUI shows these errors; the web UI should too.

## Changes

- **MessageView**: render a red error banner when `message.stopReason === "error"`, and let empty-content error messages past the empty-blocks early return.
- **ChatWindow**: keep the final assistant message renderable when it errored; treat mid-turn errored messages as displayable inside the process group.
- **useAgentSession**: raise an error notice on `message_end` when the completed assistant message has `stopReason: "error"`.

## Screenshot

Before: empty space where the assistant reply should be. After:

user bubble → assistant model label → red banner with the full `errorMessage` (monospace, pre-wrap).

image

## Test plan

- [x] `tsc --noEmit` passes
- [x] Session whose leaf is an errored assistant message shows the banner (verified in browser against a real session file)
- [x] Normal sessions (text, tool calls, process groups) render unchanged

### DIFF
--- a/components/ChatWindow.tsx
+++ b/components/ChatWindow.tsx
@@ -99,6 +99,7 @@ function countToolCalls(messages: AgentMessage[], indices: number[]): number {
 
 function hasDisplayableProcessMessage(message: AgentMessage): boolean {
   if (message.role === "assistant") {
+    if ((message as AssistantMessage).stopReason === "error") return true;
     return getDisplayableAssistantBlocks(message as AssistantMessage).length > 0;
   }
   return message.role === "custom";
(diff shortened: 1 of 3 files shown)


agegr:
Thanks for the clear report and patch!

The core issue has now been addressed in 556e2ec, including rendering empty-content assistant messages with `stopReason: "error"` and keeping final provider errors visible through the turn-grouping logic. The implementation also includes regression tests.

Thanks again for the contribution!

PR #306, opened 2026-07-30
feat: start directory picker at drive selection and add pinned sessions

Astronaut224:
- Directory picker now opens at a drive-selection view (Windows C:/D: drive letters) with a Home quick-link, instead of landing directly in the user home directory
- Session list splits into collapsible Pinned/Recent groups with a hover pin/unpin toggle, persisted to localStorage
image
image


agegr:
Thanks for the contribution. I’m going to decline this PR because the pinned/recent session list makes the sidebar behavior more complicated than I’d like, and I don’t plan to include that part.

The Windows drive selection is useful, though. I’ll implement it separately in a future change: Windows will open at the drive selection view, while macOS and Linux will continue to open at the user’s home directory by default.

agegr:
drive-selection view implemented in 248aaf464d20f6b76966b6646e743ea0904214d1

PR #315, opened 2026-07-30
fix: support wildcard enabled model refs

section2014:
## Summary
- Support wildcard patterns in `enabledModels`, such as `provider/*` and `provider/*-free`
- Preserve existing exact model matching behavior
- Keep thinking-level suffix stripping behavior

## Verification
- `node --test lib/enabled-models.test.mjs`
- `npm run lint`


section2014:
Closing as duplicate of #310, which has already been merged and covers the same enabledModels wildcard behavior more completely.

PR #325, opened 2026-07-31
feat: improve sidebar worktree and pending sessions

sleepinginsummer:
## Summary
- move new-session worktree selection into the chat input
- remove duplicate sidebar worktree selectors
- keep worktree sessions separated in the sidebar and show main checkout branches
- keep one pending unsent session per directory in the browser

## Verification
- `npm run lint`
- targeted Node tests (15 passed)


agegr:
Thanks for the implementation. After testing, I don't think Pi Web needs multi-workspace support at the moment. The scope of this change is also fairly large, so we won't merge it for now. We can revisit it when there is a clearer need.

PR #330, opened 2026-07-31
feat: add per-model token usage and cost statistics

### LINKED ISSUE #320: Feature request: per-model token usage and cost statistics
## Problem

pi-web already has the two raw ingredients for cost tracking, but nothing that aggregates them:

- **Usage data**: per-message token counts (input/output, cache read/write after #194) are shown in the chat and stored in session `.jsonl` files.
- **Pricing data**: models.dev pricing presets were integrated into the models config in #256.

What's missing is the aggregation layer: a view that answers "how much have I spent per model this week/month, and where?"

## Proposal

A usage statistics page/panel that aggregates across sessions:

- **Per model** (provider/model id): total input / output / cache-read / cache-write tokens, message count, session count.
- **Estimated cost**: tokens × the configured pricing preset (from the models.dev catalog or user-overridden prices in `models.json`); clearly marked as an estimate, hidden or marked N/A for models without pricing info.
- **Time range filtering**: today / 7d / 30d / all.
- Data source: scan session `.jsonl` files (same read path as the session list) and aggregate assistant message usage blocks — no changes to the agent runtime needed. Results can be cached server-side and invalidated on session file mtime changes.

Nice-to-haves (could be follow-ups): per-project breakdown, CSV export, daily trend chart.

## Notes

- Happy to contribute a PR for this. I'd start with a read-only `/api/usage` endpoint + a stats panel, no schema changes to session files.
- If maintainers would rather see this as an extension/plugin instead of core UI, let me know and I'll scope it that way.

ffftttft:
Closes #320

## Problem

pi-web shows per-message token counts in the chat, but nothing aggregates them across sessions — there's no way to answer "how much have I spent per model this week/month?"

## Changes

- **Usage store (`lib/usage-store.ts`)** — persists usage to `~/.pi/agent/pi-web-usage.json` as day × model buckets (input/output/cache-read/cache-write tokens, pi-recorded cost, message count, contributing session ids). Recording starts at first install (`installedAt` cutoff — no historical backfill, by design). Session files are only re-parsed when their mtime changes, per-entry seen ids make re-scans idempotent even when pi fully rewrites a session file, writes are atomic (tmp + rename), and concurrent requests share a single in-flight scan.
- **API (`GET /api/usage?range=today|7d|30d|all`)** — runs the incremental scan, then aggregates the stored buckets for the requested range: per-model totals, grand totals, and daily cost/token/message series. Invalid range values fall back to `7d`.
- **UI (`components/UsageStats.tsx`)** — modal styled after `ModelsConfig`: range tabs, summary cards (total cost / tokens / messages / sessions), a daily cost bar strip, and a per-model table with cost-share bars. Opened from a bar-chart button in the sidebar header, next to the New button.
- **i18n** — `usage.*` strings added to both `en.ts` and `zh-CN.ts`.

## Notes

- Costs reuse pi's own per-message `usage.cost` records (already priced from `models.json`), so no pricing-catalog lookup is needed and the numbers match what the chat shows per message. Entries without usage data are skipped; the UI labels costs as estimates.
- Read-only with respect to existing behavior: no agent runtime changes, no session file schema changes, no changes to existing APIs.
- Verified: `tsc --noEmit` clean, `npm run lint` clean, all existing `node --test` files pass, and an end-to-end smoke test (synthetic session file → scan → aggregate → idempotent re-scan → clean rebuild) passed. The modal was exercised against real session data in headless Chromium, including range switching.


agegr:
Thank you for the thoughtful implementation and detailed proposal.

After reviewing the scope, we've decided not to add cross-session token and cost analytics to Pi Web at this time, so we won't merge this PR.

Pi and Pi Web already expose token and cost information at the session level. A cross-session dashboard would introduce a separate accounting surface with important questions around data correctness, tracking boundaries, and the distinction between local estimates and provider/account-level usage. It also does not address subscription quota or remaining-allowance information, which is a separate concern.

For now, we want to keep Pi Web focused on the core session and agent workflow.

Thanks again for the contribution.

PR #344, opened 2026-08-01
feat: accept any file drag into the chat input

wuxiangru915:
## Summary

The chat input currently only accepts **image** drags (`useDragDrop` gates on `image/*`, and `ChatInput.addImages` filters to images). Dropping any other file silently does nothing. This PR makes the drop zone accept **any file** from the OS file manager.

## Behavior

- **Images** — unchanged: attached as base64 multimodal content and sent with the message.
- **Any other file / folder** — its real filesystem path is inserted into the input at the caret, so the agent can read it with its own tools. Browsers do not expose the real path on `File` objects, but OS-originated drags put `file://` URIs on the `dataTransfer`'s `text/uri-list`; since pi-web is a local app, that path is directly usable. Falls back to the bare file name when no path is available (e.g. drag from another browser tab).
- The full-window drop overlay now shows a document icon instead of the photo icon when the drag contains no images.

## Why path insertion (not file upload)

pi's agent core `prompt()` only supports `images` attachments, so arbitrary file content cannot be attached end-to-end. Inserting the path lets the agent fetch the file itself and works for every file type, any size.

## Implementation

- `lib/dropped-files.ts` (new): parses `text/uri-list` `file://` URIs (POSIX + Windows drive paths, URL decoding, comment/blank lines), a single-file `text/plain` fallback, and name fallback. Unit-tested in `lib/dropped-files.test.mjs`.
- `hooks/useDragDrop.ts`: accept any `kind === "file"` drag, expose `dragHasImages`, pass the `dataTransfer` to the drop callback (needed to read `text/uri-list`, only readable at drop time).
- `components/ChatInput.tsx`: new `addFiles(files, dataTransfer)` imperative method (images → existing attach flow; others → path insertion). The `insertText` logic is factored into a reusable `insertTextAtCursor`.
- `components/ChatWindow.tsx`: routes drops to `addFiles`, swaps the overlay icon based on `dragHasImages`.
- `hooks/useAgentSession.ts`: `ChatInputHandle` interface updated in sync.

## Verification

- `tsc --noEmit` clean
- `eslint` clean on changed files
- New unit tests pass (8/8); existing jiti-based tests pass, including the SSR `ChatInput.test.mjs`

### DIFF
--- a/components/ChatInput.tsx
+++ b/components/ChatInput.tsx
@@ -442,14 +442,19 @@ export const ChatInput = forwardRef<ChatInputHandle, Props>(function ChatInput({
     },
     addFiles(files: File[], dataTransfer?: DataTransfer | null) {
       if (isStreaming) return;
-      const imageFiles = files.filter((f) => f.type.startsWith("image/"));
-      const otherFiles = files.filter((f) => !f.type.startsWith("image/"));
-      if (imageFiles.length) processImageFiles(imageFiles);
-      if (!otherFiles.length) return;
+      // Resolve paths against the full file list so the index aligns with the
+      // text/uri-list entries (which include image files too).
       const uriList = dataTransfer?.getData("text/uri-list") ?? "";
       const plainText = dataTransfer?.getData("text/plain") ?? "";
-      const paths = droppedFilePaths(otherFiles, uriList, plainText);
-      const references = otherFiles.map((file, index) => droppedFileReference(file, paths[index]));
+      const paths = droppedFilePaths(files, uriList, plainText);
+      const imageFiles: File[] = [];
+      const references: string[] = [];
+      files.forEach((file, index) => {
+        if (file.type.startsWith("image/")) imageFiles.push(file);
+        else references.push(droppedFileReference(file, paths[index]));
+      });
+      if (imageFiles.length) processImageFiles(imageFiles);
+      if (!references.length) return;
       insertTextAtCursor(references.join(" "));
     },
   }));
--- a/components/ChatInput.tsx
+++ b/components/ChatInput.tsx
@@ -13,6 +13,7 @@ import {
   buildEntriesFromFiles, buildAtInsertText, extractAtQuery, filterFileEntries,
   type AtQueryMatch, type FileIndexEntry,
 } from "@/lib/file-fuzzy";
+import { droppedFilePaths, droppedFileReference } from "@/lib/dropped-files";
 import { FolderIcon, getFileIcon } from "./FileIcons";
 import { useIsMobile } from "@/hooks/useIsMobile";
 import { useI18n } from "@/hooks/useI18n";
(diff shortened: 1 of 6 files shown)


agegr:
this relies on `text/uri-list` exposing a `file://` URI, which is not reliably available for OS-originated drags. In the environment I tested, the input only receives the file name, so the agent cannot access the dropped file.

wuxiangru915:
Thanks for the review — you're right that `text/uri-list` is unreliable, and that's exactly why I've reworked the feature. The new approach (#404) **does not read browser paths at all**:

- A new **📎 file button** (and drag-and-drop) uploads the file content through the app to a persistent local directory (`~/.local/share/pi-web/attachments/<uuid>/<original-name>`) — this works identically in any browser/OS, since the browser always provides the bytes.
- The file appears as a card with its original name (uuid subdirectory per file, so names are never renamed/overwritten).
- On send, the message carries the server-side path; the locally-running agent reads the file from there.
- The UI renders attachments as clickable download cards, never showing raw paths.

So the failure mode you hit (input only receives the file name, agent can't access the file) can no longer occur. Happy to adjust anything in #404.

liuzianws:
Hi — first, this design is clean: images stay base64-attached, other files insert their real path (file:// → text/uri-list) so the agent reads them with its own tools.

I noticed this PR was closed unmerged in August, and I can't find the feature landing in a later release. I'm running v0.9.3 and I've hit this gap constantly in practice: daily I drag session logs, wiki pages and .md docs into the chat, and right now I have to fall back to typing @-filenames or telling the agent absolute paths, which is much clunkier.

For a **local** app the path-insertion approach is especially valuable — the drag source and the agent share the same filesystem, so it works for any size, any type, zero upload cost.

Would you be open to rebasing this on current main and relanding it? Happy to test a pre-release and report back (my own use case doubles as a regression test: drag .md/.log/screenshot into the input and verify each routes to the right branch).

wuxiangru915:
Hi @liuzianws, thanks for the kind words and the detailed use case — dragging `.md/.log` in daily work and falling back to typing `@` paths is exactly the pain I was trying to fix.

Quick context on history:

* #344 (insert `file://` path via `text/uri-list`) was closed because that API isn't reliable — in @agegr's environment only the filename arrives, so the agent can't access the file.
* I then reworked it as #404: upload the bytes through the app to `~/.local/share/pi-web/attachments/<uuid>/<original-name>`, no browser-path magic. That version works in any browser/OS, but was closed by design — File Explorer upload + Mention/`@` autocomplete already covers it, and the maintainer doesn't want two parallel implementations to maintain.

If you want to try it anyway, #404's Files changed / `.diff` is still viewable for reference. One heads-up: I've deleted my fork so the branch itself is gone and it was based on v0.8.7, well behind current v0.9.3 — expect manual rebase work if you apply it locally. For daily use I'd still recommend the official File Explorer + `@` flow, and upvoting an improvement to its convenience (e.g. drag-to-upload auto-mention) is probably the path most likely to land.

Happy to answer questions on the old implementation if you dig into it.

PR #369, opened 2026-08-03
fix: preserve async subagent session state

sleepinginsummer:
## Summary
- keep the session active while detached subagents are still running
- preserve user draft text when a send is not accepted during an async state transition
- cover async settlement and rejected-send paths

## Verification
- npm run lint
- node --test components/ChatInput.test.mjs hooks/useAgentSession.test.mjs


agegr (changes requested):
Thanks for working on this. I do not think this PR can be merged in its current form.

The stated fix is mixed with 12 unrelated commits covering PWA behavior, sidebar/worktree changes, session trash, notifications, and multi-skill handling. The result is 17 commits across 52 files (+3,318/-734), based on v0.8.6, with conflicts against current `main`. Current `main` also already contains newer session-lifecycle work in the same code paths.

There are several correctness issues in the final patch:

- `agent_settled` and `prompt_done` are handled as the same terminal event. An extension-injected run does not advance `promptRunId`, so settlement from the previous run can mark the new run idle and cause subsequent streaming events to be dropped. This removes the `sdkAgentActiveRef` / `rpcPromptPendingRef` protection already present on `main`.
- Every error after starting the prompt POST is treated as an accepted send. `sendAgentCommand()` also throws for explicit 4xx/5xx responses, so definitively rejected messages can still clear the draft.
- Rebuilding pending detached subagents from history ignores terminal `subagent_inspect` results. A timed-out subagent is therefore re-added before stream cleanup and can keep the SSE connection open indefinitely.
- `ChatInput` awaits `onSend()` and then clears the current input. Text or images entered while session/SSE startup is pending can be erased when the earlier send resolves.
- The unrelated PWA changes regress the existing mobile viewport behavior; `components/MobilePwaLayout.test.mjs` currently fails 3 of 4 tests.

The new tests mostly assert source-code patterns and do not exercise the React/SSE event ordering that these changes depend on. The subagent message format is also an external extension protocol, so its parsing should be isolated behind a small adapter rather than embedded throughout the shared session lifecycle.

Please recreate this change from current `main`, include only the detached-subagent behavior that is still needed, preserve the existing lifecycle protections, and add behavioral tests for event ordering, rejected HTTP responses, timed-out restoration, and draft edits during an in-flight send. A focused replacement PR will be much safer to review and merge.

PR #377, opened 2026-08-04
fix: model switching reliability, lost composer input, stranded tasks after compaction, and intranet TLS

Tom-Ma-Ming:
Four fixes found while using pi-web against a self-hosted model gateway. Each is an independent commit.

### 1. Model switching looked like a no-op and needed several tries

Three causes stacked:

- `loadSession` cleared the optimistic model unconditionally, so any reload (`agent_end`, bash settle, reconcile poll) snapped the picker back to the session file — which lags a switch, since pi appends `model_change` asynchronously and a restarted wrapper replays an older leaf.
- The dropdown skipped `set_model` entirely for the entry that already looked active, so once the label went stale there was no way to correct it: clicking the model you wanted did nothing.
- Nothing was shown between the click and the response. After the wrapper's 10-minute idle destroy, the first switch has to cold-start the agent from the session file, which takes long enough to read as "ignored".

Now the choice is applied optimistically with monotonic switch ids so a slow response cannot overwrite a newer pick, the override is dropped only once a reload agrees, every click re-issues the idempotent `set_model`, and the picker syncs from live agent state — `get_state` already returned `model`, the client just discarded it. Failures revert and surface a notice instead of only reaching the console.

### 2. Typed text was lost whenever a turn failed

`ChatInput` clears the composer right after a fire-and-forget `onSend`, and pi runs `prompt()` fire-and-forget too, so the POST returns 200 and the failure arrives later as `prompt_error`. Only `EventStreamConnectionError` was recovered. Three further gaps:

- pi writes no session file until the session has an assistant message, so a first-message failure on a fresh session lost the text outright.
- A steer/follow-up pi accepted but never delivered died with the wrapper — that queue is in-memory only, so `agent_end` reading no state silently reset it to empty.
- `restoreSubmission` read the textarea to decide how to merge, but submitting queues `setValue("")` and a fast failure can land before React flushes it, so the read saw text that was about to be wiped. This is how the recovery silently bailed on "Timed out connecting to the agent event stream".

The in-flight prompt is now kept and restored from `loadSession` only when the reloaded transcript no longer contains it, so a failure that did persist the message does not duplicate it. The queue is mirrored client-side and handed back when the agent reports no state. The composer merge uses a functional update, correct in every ordering.

### 3. Threshold auto-compaction leaves the task stranded

Upstream `_checkCompaction` documents "Threshold: Context over threshold, compact, NO auto-retry (user continues manually)" and returns `hasQueuedMessages()`, so with an empty queue the agent loop just ends mid-task and the user has to nudge it by hand.

**This commit deliberately departs from that behaviour** — worth a maintainer opinion, and easy to drop if you would rather pi-web match the CLI. pi-web now sends the nudge itself, armed only for threshold compaction (`reason !== "manual"`, `!willRetry`, since the overflow path resumes on its own). Bounded to three consecutive continuations so a task that keeps overflowing cannot quietly spend tokens in a loop; any user message resets the streak.

### 4. Intranet model endpoints reached by IP fail before the request leaves the process

A provider at `https://<ip>:<port>` serving a certificate whose SAN only lists a DNS name is rejected with `ERR_TLS_CERT_ALTNAME_INVALID`, and the UI shows nothing at all — no response, no error. `curl -k` succeeds, so the endpoint looks healthy.

`lib/tls-overrides.ts` probes such origins found in `models.json`, reuses the certificate's SAN name as the TLS servername so the chain and hostname are still fully verified, and falls back to skipping verification for that one origin only if the probe fails. It resolves per-origin connect options that the existing dispatcher's client and pool factories merge in, rather than competing over `setGlobalDispatcher`. Opt-out and manual overrides via `PI_WEB_TLS_AUTO=0`, `PI_WEB_TLS_SERVERNAME_OVERRIDES`, `PI_WEB_TLS_INSECURE_HOSTS`.

Verified through the real global fetch path: the intranet endpoint answers 200, and `https://self-signed.badssl.com` is still rejected with `DEPTH_ZERO_SELF_SIGNED_CERT`, so public verification is not weakened.

### Checks

`tsc --noEmit`, `eslint .`, 240/240 tests, and `next build` all pass on top of v0.8.6.

No automated regression tests were added: these are event-ordering and TLS paths, and the existing suite renders statically with no DOM harness available. Manual verification covered each fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/components/ChatInput.tsx
+++ b/components/ChatInput.tsx
@@ -37,13 +37,14 @@ interface Props {
   onPromptWithStreamingBehavior?: (message: string, behavior: "steer" | "followUp", images?: AttachedImage[]) => void;
   isStreaming: boolean;
   model?: { provider: string; modelId: string } | null;
-  isAutoModelSelection?: boolean;
   modelNames?: Record<string, string>;
   modelList?: { id: string; name: string; provider: string }[];
   modelError?: string | null;
   /** Diagnostics from resolving `enabledModels`, e.g. a pattern that matched nothing. */
   modelScopeWarnings?: string[];
   onModelChange?: (provider: string, modelId: string) => void;
+  /** A switch is in flight — shown on the picker so the click never looks ignored. */
+  modelSwitching?: boolean;
   onCompact?: () => void;
   onAbortCompaction?: () => void;
   isCompacting?: boolean;
@@ -76,6 +77,7 @@ export interface ChatInputHandle {
   insertIfEmpty: (text: string) => void;
   prependText: (text: string) => void;
   addImages: (files: File[]) => void;
+  restoreSubmission: (text: string, images?: ChatDraftImage[]) => void;
 }
 
 const TOOL_PRESETS = ["off", "default", "full"] as const;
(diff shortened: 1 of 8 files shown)


agegr:
Closing this PR without merging.

The valid issues identified here are not yet fixed on main. They will be re-evaluated and addressed in separate, focused pull requests based on the current main branch.

Validation found that the model-switching root cause described in this PR does not apply to Pi 0.83.0: model changes are written synchronously before the set_model request returns. The remaining changes also need revised implementations and dedicated regression tests, particularly around composer recovery, compaction event ordering, and TLS security.

This PR will remain available as investigation context, but its implementation will not be merged as-is.

PR #382, opened 2026-08-04
perf: stream agent message deltas instead of full snapshots (#375)

### LINKED ISSUE #375: 远程访问时带宽放大 ~150 倍：events 流每次更新都重发完整消息
## 问题概述

当 **远程访问** pi-web 时（通过 VPN / 公网），带宽消耗极大：events 流在每次流式更新时都会**重发已累积的完整消息**。实测放大倍数约 **150 倍**：模型输出 ~15 KB 内容，实际传输 ~2.2 MB；`thinking: high` 时可达 6 MB 以上。

## 环境

- pi-web 0.8.6，`@earendil-works/pi-coding-agent` 0.83.0
- 浏览器客户端通过 VPN 隧道访问（高延迟 / 按流量计费的网络下尤其明显）

## 根因

RPC 协议中每条 `message_update` 事件都携带**完整的 message 对象**（已生成的全部 text/thinking 累积内容）**以及** `assistantMessageEvent` delta（见 `docs/rpc.md`）：

```json
{"type":"message_update","message":{...完整累积消息...},"assistantMessageEvent":{"type":"text_delta","delta":"Hello"}}
```

pi-web 的 events 路由（`.next/server/app/api/agent/[id]/events/route.js`）把这些事件**几乎原样**转发给浏览器——只删掉了 `assistantMessageEvent` 字段：

```js
let j = new Set(["turn_start","turn_end","tool_execution_update"]);
// ...
if ("message_update" === a.type) {
  let b = { ...a };
  delete b.assistantMessageEvent;   // 保留了完整的 message
  return b;
}
```

前端（`.next/static/chunks/app/page-*.js`）则在每次 `message_update` 时重新渲染整条消息。

由于流式输出大约每 50 字符产生一个更新块，**每个块都重发此前生成的全部内容** → O(n²) 传输：

| 场景（回答 10,000 字符） | 实际发送到浏览器的字节 | 相对内容的放大倍数 |
|---|---|---|
| thinking off | ~1.0 MB | ~105x |
| thinking low | ~1.4 MB | ~120x |
| thinking high | ~6.2 MB | ~255x |
| 理想增量传输（仅 delta） | ~15 KB | 1x |

本机访问无感；远程通过 VPN / 按流量计费的网络使用时问题很明显。

## 建议修复方向（任一均可）

1. **服务端节流合并（最简单，无需改前端）**：对 `message_update` 事件做短时间窗口缓冲（如 50–100 ms），每个窗口只发最新一条完整 message。不用动前端即可消除 O(n²) 曲线。
2. **前端增量渲染**：用 `assistantMessageEvent` 的 delta 增量拼接，而不是每次事件都重新渲染完整消息。
3. **两者结合**：服务端对突发流做合并，前端做 delta 追加渲染，兼顾平滑度。

如果维护者认可方案 1，我可以提交对应 PR。

Naoki326:
Fixes #375 — remote access bandwidth amplification (~150x) caused by the events stream resending the full accumulated message on every `message_update`.

## Root cause
The events route forwarded each SDK `message_update` (which carries the complete accumulated message plus an `assistantMessageEvent` delta) almost verbatim — only stripping the delta field. Since streaming emits a chunk roughly every 50 characters, every chunk re-sent everything generated so far → O(n²) transfer (~15 KB response cost ~2.2 MB over the wire).

## Change
- **Server** (`app/api/agent/[id]/events/route.ts`): `message_update` is projected onto a light `message_delta` event — only the slim delta fields (`contentIndex`/`delta`/`content`/`toolCall`) are forwarded; `partial` (the full accumulated message) is stripped. `done`/`error` are omitted (their `reason` already arrives in `message_end`'s full message). Unknown/missing delta types degrade to the existing full-snapshot fallback. On SSE (re)connect, the current partial message is injected as a snapshot so mid-stream page reloads recover immediately instead of waiting for `message_end`.
- **Client** (`hooks/useAgentSession.ts` + new `lib/stream-delta.ts`): streamed messages are assembled locally by appending deltas (`text`/`thinking`/`toolCall`), keeping the existing snapshot-overwrite fallback. Rendering is unchanged.
- `toolcall_start` carries `contentIndex` only; `id`/`name` are extracted from `partial` (two fields only, not the whole message) so tool cards show the tool name while streaming.

## Verification
- typecheck + lint clean; full test suite: 261 tests (250 pass; 11 pre-existing environment failures identical to HEAD).
- New unit tests for delta assembly (`lib/stream-delta.test.mjs`) plus projection assertions in `app/api/agent/events-route.test.mjs`.
- Simulation: ~13.5 KB response previously cost ~1.8 MB over the wire (~139x amplification); now ~55 KB (~4x constant framing overhead — O(n) instead of O(n²)).

### DIFF
--- a/app/api/agent/[id]/events/route.ts
+++ b/app/api/agent/[id]/events/route.ts
@@ -5,9 +5,52 @@ export const dynamic = "force-dynamic";
 
 const OMITTED_EVENT_TYPES = new Set(["turn_start", "turn_end", "tool_execution_update"]);
 
+// assistantMessageEvent 中可安全转发给浏览器的轻量字段。`partial`（完整累积
+// 消息）被有意剥离——转发它会重新引入 O(n²) 传输（每个 delta 都重发此前
+// 生成的全部内容）。`done`/`error` 不在此列：它们的 reason 已包含在
+// message_end 的完整消息中，前端从 message_end 获取即可。
+const DELTA_FIELDS: Record<string, readonly string[]> = {
+  start: [],
+  text_start: ["contentIndex"],
+  text_delta: ["contentIndex", "delta"],
+  text_end: ["contentIndex", "content"],
+  thinking_start: ["contentIndex"],
+  thinking_delta: ["contentIndex", "delta"],
+  thinking_end: ["contentIndex", "content"],
+  toolcall_start: ["contentIndex"],
+  toolcall_delta: ["contentIndex", "delta"],
+  toolcall_end: ["contentIndex", "toolCall"],
+};
+
+// done/error 不转发：reason 已随 message_end 的完整消息下发，前端无需重复接收。
+const OMITTED_DELTA_TYPES = new Set(["done", "error"]);
+
 function toClientEvent(event: AgentEvent): AgentEvent | null {
   if (OMITTED_EVENT_TYPES.has(event.type)) return null;
   if (event.type === "message_update") {
+    const delta = event.assistantMessageEvent as ({ type: string } & Record<string, unknown>) | undefined;
+    const fields = delta && typeof delta.type === "string" ? DELTA_FIELDS[delta.type] : undefined;
+    if (delta && fields) {
+      // 增量事件：只转发轻量 delta 字段，浏览器端据此拼接流式消息。
+      const slimDelta: Record<string, unknown> = { type: delta.type };
+      for (const field of fields) slimDelta[field] = delta[field];
+      // toolcall_start 只携带 contentIndex；id/name 仅存在于 partial（完整累积
+      // 消息）中。只提取这两个字段注入，不转发整个 partial，避免重新引入
+      // O(n²) 传输——否则流式期间工具卡片无名无 id。
+      if (delta.type === "toolcall_start") {
+        const partial = delta.partial as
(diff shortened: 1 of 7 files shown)


agegr:
Thanks for the detailed investigation, implementation, tests, and bandwidth measurements. The diagnosis is correct, and moving `message_update` to delta-only transport is the right direction.

Since this PR was opened, pi-web has upgraded to Pi 0.84. Pi now defines an official delta-only JSON/RPC wire contract:

- `message_start` provides the initial full message
- `message_update` keeps its existing name and carries only `assistantMessageEvent` without `partial`
- `message_end.message` is the authoritative final state

We want pi-web to follow that contract directly instead of introducing a separate `message_delta` event and maintaining an additional streaming protocol. The current implementation also adds parallel client state, custom delta field whitelisting, tool-call special cases, and snapshot fallback behavior that would make future Pi upgrades harder to reason about.

We will implement a smaller replacement that uses the official event types, keeps a single client-side partial-message reducer, treats `message_end` as authoritative, and sends one full baseline when reconnecting.

Therefore, I’m closing this PR rather than asking for a rebase or further changes. The work here, especially the measurements and reconnect analysis, was very useful in deciding the final direction. Thank you for contributing.

PR #393, opened 2026-08-05
perf: coalesce streamed message_update events to cut remote bandwidth (#375)

### LINKED ISSUE #375: 远程访问时带宽放大 ~150 倍：events 流每次更新都重发完整消息

shani-singh1:
## Summary

Fixes #375.

On remote access (VPN / metered / public network), pi-web's SSE stream is hugely bandwidth-heavy: every `message_update` event carries the **full accumulated message**, and the agent emits one per streamed chunk. The events route forwarded each one, so the browser received the entire message again on every chunk — transfer grows **O(n²)** with message size. The reporter measured **~150×** amplification over the actual content (~15 KB content → ~2.2 MB transferred; >6 MB with `thinking: high`).

## Fix (server-side only)

Coalesce streamed `message_update` events in the events route:

- Keep only the **latest** pending `message_update` (each already carries the full message, so older ones are redundant) and flush it on a short **~80 ms** timer — a burst of chunks collapses into one send, while streaming stays smooth (~12/s).
- **Any other event** (tool / turn / `agent_end`) flushes the pending update **first**, so a `message_update` is never reordered past a later event, and the message's final content is always delivered before the next boundary event.
- The flush timer is cleared on client disconnect.

No protocol or frontend change: the client still renders each `message_update` it receives exactly as before — it just receives far fewer of them, which also reduces client re-render cost. Full delta rendering (frontend appends `assistantMessageEvent` deltas) would take this to 1× and is a natural follow-up; this change is the contained, low-risk server-side win.

The coalescing logic is extracted into `lib/event-coalescer.ts` so it can be unit-tested independently of the SSE plumbing.

## Testing

- `tsc --noEmit` passes
- `eslint` passes
- `node --test lib/event-coalescer.test.mjs` — 5/5 pass, covering: buffering until flush, coalescing consecutive updates to the latest, flushing pending before a following non-update event (ordering), non-update passthrough, and no-op flush.

### DIFF
--- a/app/api/agent/[id]/events/route.ts
+++ b/app/api/agent/[id]/events/route.ts
@@ -1,7 +1,14 @@
 import { resolveSessionPath } from "@/lib/session-reader";
 import { getRpcSession, startRpcSession, type AgentEvent } from "@/lib/rpc-manager";
+import { createMessageUpdateCoalescer } from "@/lib/event-coalescer";
 
 export const dynamic = "force-dynamic";
+
+// Buffer window for coalescing streamed message_update events. Each update
+// carries the full accumulated message, so forwarding every one amplifies
+// transfer O(n^2) (#375). ~80ms keeps streaming visibly smooth (~12/s) while
+// collapsing bursts of updates into a single send.
+const MESSAGE_UPDATE_FLUSH_MS = 80;
 
 const OMITTED_EVENT_TYPES = new Set(["turn_start", "turn_end", "tool_execution_update"]);
 
@@ -48,9 +55,23 @@ export async function GET(
       // Send initial connected event
       encode({ type: "connected", sessionId: id });
 
+      // Coalesce streamed message_update events so remote clients don't receive
+      // the full accumulated message on every chunk (#375).
+      const coalescer = createMessageUpdateCoalescer(encode);
+      let flushTimer: ReturnType<typeof setTimeout> | null = null;
+      const scheduleFlush = () => {
+        if (flushTimer) return;
+        flushTimer = setTimeout(() => {
+          flushTimer = null;
+          coalescer.flush();
+        }, MESSAGE_UPDATE_FLUSH_MS);
+      };
+
       const unsubscribe = session.onEvent((event) => {
         const clientEvent = toClientEvent(event);
-        if (clientEvent) encode(clientEvent);
+        if (!clientEvent) return;
+        coalescer.push(clientEvent);
+        if (coalescer.hasPending()) scheduleFlush();
       });
 
       // Heartbeat every 30s to prevent server/proxy timeout (Next.js default ~120-150s)
@@ -65,6 +86,7 @@ export async function GET(
       // Cleanup when client disconnects
       const cleanup = () => {
         clearInterval(heartbeat);
+        if (flushTimer) clearTimeout(flushTimer);
         unsubscribe();
         controller.close();
       };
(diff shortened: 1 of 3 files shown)


agegr:
Thanks for the contribution and for keeping the proposed change server-side and contained. Coalescing full `message_update` snapshots is a reasonable short-term mitigation and has a smaller synchronization risk than introducing a custom client protocol.

However, a fixed 80 ms window only reduces the constant factor. It still sends a growing full-message snapshot repeatedly, so total transfer remains O(n²). When update intervals are 80 ms or slower, little or no coalescing occurs. Because of that, this cannot be the final fix for #375.

pi-web has now upgraded to Pi 0.84, which defines an official delta-only JSON/RPC contract for `message_update`. We will align the SSE transport with that contract instead:

- keep the `message_update` event name
- forward only `assistantMessageEvent` without `partial`
- use `message_start` as the baseline
- treat `message_end.message` as the authoritative final state
- send one full baseline on reconnect

This provides linear wire usage without maintaining repeated full snapshots. I’m therefore closing this PR in favor of the replacement implementation. We will keep #375 open until that implementation lands.

Thank you for the clear implementation and tests.

PR #413, opened 2026-08-07
feat: make the chat input manually resizable & maximize / restore button

isWittHere:
## Summary

Allows giving the chat input box a fixed height instead of relying only on
content-driven auto-growth. Two ways to resize:

1. **Drag handle** on the composer's top edge — drag up to grow, down to
   shrink; double-click or press Enter to reset. The height is persisted to
   localStorage and clamped to a viewport-aware ceiling on resize.
image

2. **Maximize / restore button** next to the send button — jumps to the
   ceiling, or back to the auto height. It also appears while the agent runs,
   next to the steer/follow-up queue buttons.
image
image

Entering/leaving manual mode syncs the textarea height so flex stretch takes
over cleanly, and the manual floor matches the composer's natural single-line
height so grabbing the handle from auto mode transitions without a jump.

## Changes

**New:**
- `hooks/useResizableHeight.ts` — vertical counterpart of `useResizablePanel`:
  pointer drag with a 3px dead zone, keyboard support (↑/↓/Home/End/Enter),
  localStorage persistence, viewport re-clamp, and drag cancellation on blur /
  tab hide.

**Edited:**
- `components/ChatInput.tsx` — wire the handle to the composer shell; suspend
  auto-growth in manual mode via a shared `applyAutoHeight` helper; add the
  maximize/restore button (shared `maximizeButton` element present in both the
  idle and streaming branches).
- `app/globals.css` — resize-handle styles (hover-only highlight line).
- `lib/i18n/messages/{en,zh-CN}.ts` — `layout.resizeInput` / `maximizeInput` /
  `restoreInputHeight`.

## Testing

- `node_modules/.bin/tsc --noEmit` — clean
- `npm run lint` — clean
- `node --test` — no new failures (pre-existing baseline failures unchanged)


agegr:
Thanks for the thoughtful implementation and for paying attention to details like keyboard support, viewport clamping, and persistence.
After reviewing the feature, I don't think manual resizing and maximizing belong in pi-web. The current content-driven auto-grow behavior already covers the common case well, while a resize handle, a persistent manual-height mode, and another control beside the primary chat actions add UI and maintenance complexity for a relatively narrow use case.
I'd like to keep pi-web's composer simple and predictable, so we won't be merging this PR. This is a product-scope decision rather than a concern with the quality of the implementation. Thanks again for taking the time to build it.

PR #442, opened 2026-08-09
Add provider availability toggles

ThePig005:
## Summary

- Add per-provider enable/disable toggles to the Models configuration panel.
- Keep disabled custom providers visible and fully configured while excluding them from the chat model list and new-session model scope.
- Apply the same availability filter to HTTP model listing and RPC session startup.
- Keep UI state in a separate `models-ui.json` sidecar because pi core validates the `models.json` schema and rejects unknown provider fields.

## Compatibility

- Existing `models.json` files remain unchanged and require no migration.
- Missing, malformed, or empty UI state is treated as no disabled providers.
- Provider rename and deletion keep the sidecar state consistent.
- This change targets custom providers configured in `models.json`; managed OAuth/API-key provider rows remain on their existing paths.

## Tests

- `npm run lint`
- `node --test lib/model-scope.test.mjs lib/provider-availability.test.mjs lib/provider-listing.test.mjs`

All 24 related tests pass.


agegr:
Thanks for the contribution. After reviewing it, I think this adds more complexity than the feature needs. We want to keep Pi Web simple, so I’ll close this PR for now. Thanks again!

PR #448, opened 2026-08-09
feat: offline context usage estimation for sessions

111-test-111:
## What

Estimates a session's context-window usage (percent + token counts) by reading the session `.jsonl` file directly — no live AgentSession required — and serves it from the session API routes:

- `GET /api/sessions/[id]` and `GET /api/sessions/[id]/context?leafId=` responses gain a best-effort `contextUsage` field (failures never fail the response)
- The hook consumes file-derived `contextUsage` on session/context load, so non-running sessions show usage immediately; the system-prompt panel is loaded on demand via `get_state` and refreshes live usage
- Extracts the inline model loading from `/api/models` into `lib/models-loader.ts` (shared, adds `contextWindow` to `modelList`) since the estimator needs model context windows

## Notes

- Purely additive to server responses; no UI behavior changes for running sessions.
- 9 new tests (7 lib + 1 hook + 1 route-wiring). Full suite: 432 pass / 0 fail; `tsc --noEmit` and eslint clean.
- Heads-up: this touches `app/api/sessions/[id]/route.ts` imports/response, as does the companion branch-labels PR — the second one to land may need a trivial rebase.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,10 +70,12 @@ app/api/
 
 lib/
   agent-client.ts      typed fetch helper for /api/agent commands
+  context-usage.ts     offline context-usage estimation from session files
   draft-store.ts       local draft persistence helpers
   file-access.ts       allowed file roots for /api/files and worktrees
   file-paths.ts        client/server path encoding helpers
   markdown.ts          shared markdown helpers
+  models-loader.ts     loadModels() — models registry loader shared by /api/models and session routes
   npx.ts               npx runner used by skill install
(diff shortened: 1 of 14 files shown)


agegr:
Offline context usage should use a lightweight calculation path. The current implementation calls `createAgentSessionServices()` and loads models and extensions before returning the session response, so a slow extension can block session history loading.

Please resolve the model's `contextWindow` without initializing full agent services or project extensions. If the lightweight lookup fails, return `contextUsage: null` without delaying the session response.

PR #458, opened 2026-08-10
feat(sidebar): group sessions into display-only categories

k0n9-devx:
This adds an optional, display-only folder/category layer on top of the
flat session list in the sidebar. Users can create, rename, and delete
categories (up to arbitrary nesting depth) and assign sessions to them
via a dropdown menu on each session row.

### Design: pure presentation layer

Categories are stored in `~/.pi/agent/session-folders.json` (same
directory and atomic-write pattern as `models.json`). **No session
`.jsonl` file is ever read or written by this feature.** The
`SessionManager`, RPC/agent logic, and the session persistence format
are completely untouched.

Consequence: deleting a category **never deletes a session**. Direct
child sessions become unfiled; direct child subcategories are promoted
to the deleted category's parent.

When `session-folders.json` is absent or deleted, the UI behaves
exactly as before — all sessions appear flat. The feature is trivially
revertible.

### Refactor note

`SessionItem` / `SessionTreeItem` were extracted from
`SessionSidebar.tsx` into a shared `components/SessionTreeItem.tsx`
so both the original sidebar and the new `SessionFolderTree` can render
sessions with the same rename/delete interactions. Extracted code is
unchanged apart from the new optional "move to category" action.

### Testing

- `lib/session-folders.test.mjs`: 15 cases covering CRUD,
  non-destructive delete/promotion, assignment, and pruning.
- Existing `SessionSidebar.test.mjs` source-inspection tests split
  between the two files where the code now lives.
- All 441 existing tests pass.

### i18n

Both `en` and `zh-CN` strings added (category → "分类" in Chinese).


agegr:
Thanks for the contribution. I’m going to decline this PR because the category system adds significant persistence, API, and UI complexity for an optional feature. I’d prefer to keep pi-web’s session model and sidebar simple.

PR #464, opened 2026-08-10
feat: add manual file editing and file/folder management

AxelWT:
Add env-gated (NEXT_PUBLIC_PI_WEB_FILE_EDITING) file editing to FileViewer and create/delete operations to FileExplorer.

FileViewer:
- Edit mode toggle (pencil): textarea with line numbers, Tab-to-indent, Cmd/Ctrl+S to save, Esc to cancel
- PATCH /api/files/[...path] reuses upload route's allow-list + double-realpath security; temp file + atomic rename; chmod temp to original mode so executables keep their bit
- Binary files disable Edit; SSE change events set external-modified banner instead of clobbering the textarea; beforeunload guards drafts
- saveEdit refreshes git diff marker and file tree; reload confirms before overwriting unsaved edits

FileExplorer:
- POST ?type=create: empty file (flag:wx, EEXIST→409) or directory inside realpath-hardened parent
- DELETE: lstat to detect symlinks — remove the link itself, never its target; allowed roots themselves are never deletable
- AppShell closes tabs pointing at deleted files/directories

Misc:
- Basic Auth username from PI_WEB_USERNAME (defaults to "pi")
- .env.example documents the editing toggle and auth/host vars
- i18n keys added (en + zh-CN)


AxelWT:
截屏2026-08-10 18 03 19

新增一个 创建 删除和编辑文件的功能，支持自己手动编辑文件  引入一个环境变量作为开关来控制

agegr:
Thanks for the contribution! This PR significantly expands Pi Web’s scope and maintenance cost by adding file editing, creation, and deletion. We’d like to keep Pi Web focused on agent interaction, so we won’t merge this for now.

PR #470, opened 2026-08-11
feat: add MCP server management to the plugins panel

luckyyzh:
# Add MCP server management to the Plugins panel

## Summary

Adds a full MCP (Model Context Protocol) server management section to the existing
**Plugins** config panel, so users can view, configure, enable/disable, test, move and
delete MCP servers without editing JSON files by hand.

- New backend route `GET|POST /api/mcp` to read and mutate the MCP configuration
  (project `.pi/mcp.json` + global `~/.pi/agent/mcp.json`), including a stdio/HTTP
  connectivity test.
- New left-sidebar "MCP 服务器" section in `PluginsConfig` (grouped by project/global
  scope), with an "添加MCP" button mirroring the existing "添加插件" button.
- A right-hand detail pane (enable/disable toggle, command/URL, args, env keys
  redacted, options, source file, test/edit/move/delete actions) and an add/edit form
  with both a guided mode and an advanced JSON mode.

## Motivation

Users already manage pi plugins from the web UI. MCP server configuration currently
requires editing `mcp.json` by hand (or via a chat slash command). Bringing it into the
same panel gives a consistent, discoverable management surface and parity with the
plugin experience.

## What changed

| File | Change |
| --- | --- |
| `app/api/mcp/route.ts` | New route: list (merged global+project), add/update/remove/enable/disable/move, fetch raw def, and `initialize`+`tools/list` connectivity test for stdio & HTTP servers |
| `lib/api-types.ts` | New `McpScope`, `McpServerInfo`, `McpResponse` types |
| `lib/i18n/messages/en.ts` / `lib/i18n/messages/zh-CN.ts` | New `mcp.*` translation keys for the MCP panel (en + zh-CN) |
| `components/PluginsConfig.tsx` | MCP section in the left list, `McpServerDetail`, `AddMcpServer` (basic + JSON modes), plus load/save/test/move handlers |

## How it works

- The route reads the same files pi-mcp-adapter uses: project `.pi/mcp.json` and global
  `~/.pi/agent/mcp.json` (`{ mcpServers, settings }`). Global is the base, project
  overrides. Writes go to the chosen scope.
- Project-scope writes require the project to be trusted (same rule as plugins).
- Env values are shown as key names only in the detail view; the advanced JSON editor
  fetches the full raw definition (including env values) only when editing.
- A test spawns the stdio server and runs `initialize` → `tools/list` (HTTP servers use a
  POST `initialize`).

## Known limitations / open questions

- All new UI strings are localized through `lib/i18n` (en + zh-CN).
- The stdio connectivity test spawns with `shell: true` on the configured command (same
  assumption as the underlying MCP tooling); no shell injection is exposed to users since
  the command comes from the user's own config.
- A couple of UI labels borrow wording from the plugin panel; feel free to bikeshed naming
  (e.g. the section title).
- The advanced JSON editor returns the full raw definition (including env values) for the
  selected server when editing. This intentionally exposes secrets in the editor; it is the
  same config the user already owns locally.

## Testing

- Type check clean (`tsc --noEmit`), production build succeeds (`next build --webpack`).
- `GET /api/mcp` returns configured servers with correct scope/source and redacted env.
- POST actions verified end-to-end: add, disable, move scope, remove, test
  (returns `github-mcp-server 0.6.2 · 26 个工具`), and `get` returns the full raw def.
- The panel was exercised against a real project config (github server) and a throwaway
  server, with the project file left unchanged afterwards.

## Notes for reviewers

The change is intentionally localized to 3 files. The feature is additive — it only
registers a new route and extends the existing plugins panel with a section — so it should
be low-risk to existing behavior.


agegr:
Thanks for the contribution! We’re currently refactoring the settings panel, so we won’t merge this PR for now. MCP support will be added later after the refactor is complete.

PR #498, opened 2026-08-14
feat: show reply timing summary on completed AI replies

molicherry:
## Summary
- Adds a one-line summary under the final assistant message of each turn,
  showing the end time, total turn duration, time-to-first-token, and average
  output speed.
- Duration is computed from the session entry's persisted timestamp (generation
  end) minus the turn's opening user message. The message's own `timestamp`
  field is actually when generation *started*, which previously rounded the
  duration to ~0s.
- Time-to-first-token is measured server-side in `AgentSessionWrapper` and
  stamped onto the message before pi persists it, so it survives page reloads.
- Average speed aggregates output tokens across the turn's assistant messages;
  rates below 10 tok/s show one decimal so a positive rate is never hidden.
- Adds zh-CN / en i18n keys.

## Files changed
- `components/ChatWindow.tsx`: computes `turnStartTimestamp` / `firstTokenMs` /
  `turnOutputTokens` per turn
- `components/MessageView.tsx`: renders the summary line + duration formatting
- `lib/rpc-manager.ts`: measures and persists time-to-first-token server-side
- `lib/session-reader.ts`: attaches the entry `endedAt` timestamp on load
- `lib/types.ts`: adds `timeToFirstTokenMs` / `endedAt` to `AssistantMessage`
- `lib/i18n/messages/en.ts` / `zh-CN.ts`: summary labels

## Test plan
- `tsc --noEmit` passes
- `npm run lint` passes (no new warnings)
- `node --test` passes for the affected suites (MessageView, streaming-message,
  normalize, agent-event-wire, useAgentSession)

## Notes
- `time-to-first-token` only appears for turns created after this change;
  historical sessions never recorded it, so it can't be back-filled.


agegr:
Thanks for putting this together. After consideration, we don’t think these timing metrics provide enough value in the main chat interface to justify the additional visual complexity. The persistent summary line makes replies feel cluttered, while TTFT, token speed, and tool duration are primarily diagnostic metrics rather than information most users need during normal conversations.We’d prefer to keep the chat UI focused, so we won’t be merging this PR.

Issue #329, opened 2026-07-31
"Untrusted API request" (403) on localhost: sandboxed iframe preview (Origin: null + cross-site) and loopback spelling mismatch


drlhw:
# Issue 2: "Untrusted API request" (403) in pure localhost usage when origin/host use different loopback spellings

## Summary

When using pi-web locally, some API requests fail with `{"error":"Untrusted API request"}` (HTTP 403) even though the browser is on `localhost` — because the request `Origin` uses a *different loopback spelling* (`127.0.0.1` or `::1`) than the request `Host`.

This is **not** a reverse-proxy scenario (see #286) and **not** the PWA session-export case (see #301) — it happens in a plain local browser session.

## Environment

- pi-web version: 0.8.2
- pi version: 0.82.1
- OS: macOS
- Browser: Safari (also reproducible via curl)

## Steps to reproduce

Open pi-web at `http://localhost:30141` (page origin = `http://localhost:30141`), then trigger any request whose `Host`/`Origin` pair uses different loopback spellings:

```bash
# Simulates a page opened via localhost making a request with origin 127.0.0.1
curl -s "http://localhost:30141/api/files/.../file.docx?type=preview" \
  -H "Host: localhost:30141" \
  -H "Origin: http://127.0.0.1:30141"
# → HTTP 403 {"error":"Untrusted API request"}

# Works fine when both match:
curl -s "...?type=preview" -H "Host: localhost:30141" -H "Origin: http://localhost:30141"
# → HTTP 200
```

## Root cause

The request-security middleware (`lib/request-security.ts`) requires `Origin` to **exactly equal** the request's own `Host`:

```ts
const originOk = !hasOriginHeader || (
  secFetchSite !== "cross-site"
  && (!originHeader || getRequestOrigin(request) === canonicalOrigin(originHeader))
);
```

`localhost`, `127.0.0.1`, and `::1` are all the same machine, but they compare unequal as strings — so browsers (or iframes, extensions, PWA flows) that normalize `Host` to one spelling while keeping the page `Origin` in another spelling get rejected.

## Suggested fix

Treat loopback origins as mutually equivalent: if the request `Origin` hostname is `localhost`, `127.0.0.1`, or `::1`, allow it (the request `Host` is already validated to be loopback / allowed-list, which is the actual DNS-rebinding protection):

```ts
if (originHeader) {
  const originHost = new URL(originHeader).hostname;
[...]

PR #511, opened 2026-08-16
fix: accept same-origin requests behind a scheme-rewriting proxy

### LINKED ISSUE #497: 403 "Untrusted API request" behind a proxy that rewrites the Origin scheme (Azure Dev Tunnels)
## Summary

Behind an Azure Dev Tunnel, pi-web renders existing conversations fine, but **every request carrying an `Origin` header — i.e. every POST, so sending an instruction — is rejected**:

```
403 {"error":"Untrusted API request"}
```

`isApiRequestOriginAllowed()` in `lib/request-security.ts` requires the `Origin` header to equal `` `${new URL(request.url).protocol}//${host}` ``. Next.js derives the URL protocol from `x-forwarded-proto`, but the Dev Tunnel relay forwards `x-forwarded-proto: https` while rewriting `Origin` to **`http://`**. The two disagree on the scheme alone, so a genuinely same-origin request is rejected.

Still present on `main` (`lib/request-security.ts`, unchanged since `d362764`).

## Environment

| | |
| --- | --- |
| `@agegr/pi-web` | 0.8.8 (latest) + `main` @ HEAD |
| Next.js | 16.2.12 |
| Node.js | 24 |
| OS | Linux, systemd user service |
| Launch | `PI_WEB_HOSTNAME=127.0.0.1`, port 30141, published with `devtunnel host` |

## What the relay actually sends

Captured by temporarily replacing pi-web with a header-echo server on the same loopback port, then issuing an authenticated POST to the tunnel URL:

```json
{
  "method": "POST",
  "url": "/api/agent/new",
  "headers": {
    "host": "localhost:30141",
    "origin": "http://localhost:30141",
    "x-forwarded-proto": "https",
    "x-forwarded-host": "<tunnel-name>-30141.<region>.devtunnels.ms",
    "x-forwarded-scheme": "https",
    "x-scheme": "https",
    "x-forwarded-port": "443",
    "sec-fetch-site": "same-origin"
  }
}
```

The relay normalizes both `Host` and `Origin` onto `localhost:30141` and reports the real scheme out-of-band. Note it downgrades **only** the `Origin` scheme; `x-forwarded-proto` stays `https`.

## Why it fails

```ts
// lib/request-security.ts
function getRequestOrigin(request: Request): string | null {
  const requestUrl = new URL(request.url);            // protocol: "https:" (from x-forwarded-proto)
  const host = request.headers.get("host");           // "localhost:30141"
  return host ? canonicalOrigin(`${requestUrl.protocol}//${host}`) : null;
}                                                     // → "https://localhost:30141"

export function isApiRequestOriginAllowed(request: Request): boolean {
  ...
  const requestOrigin = getRequestOrigin(request);
  return requestOrigin !== null && canonicalOrigin(origin) === requestOrigin;
}                                                     // "http://localhost:30141" !== "https://localhost:30141" → false
```

This matches the symptom precisely: browsers omit `Origin` on same-origin GETs, so navigation and GET fetches pass, while every POST is blocked. `isApiRequestHostAllowed()` is not involved — the relay's `Host` is already loopback — so `PI_WEB_ALLOWED_HOSTS` (the remedy from #286) cannot help here.

## Reproduction

No tunnel required; with pi-web on `127.0.0.1:30141`:

```bash
# Blocked (403) — exactly what the relay sends
curl -i -X POST http://127.0.0.1:30141/api/agent/new \
  -H 'Host: localhost:30141' \
  -H 'Origin: http://localhost:30141' \
  -H 'X-Forwarded-Proto: https' \
  -H 'Sec-Fetch-Site: same-origin' \
  -H 'Content-Type: application/json' -d '{}'
# HTTP/1.1 403  {"error":"Untrusted API request"}

# Same request without x-forwarded-proto → guard passes (400 "cwd is required")
curl -i -X POST http://127.0.0.1:30141/api/agent/new \
  -H 'Host: localhost:30141' \
  -H 'Origin: http://localhost:30141' \
  -H 'Sec-Fetch-Site: same-origin' \
  -H 'Content-Type: application/json' -d '{}'
```

## Suggested fix

**Option A — compare hosts, not full origins** (preferred; also fixes the loopback-spelling family of reports):

```ts
export function isApiRequestOriginAllowed(request: Request): boolean {
  const origin = request.headers.get("origin");
  const fetchSite = request.headers.get("sec-fetch-site");
  if (fetchSite === "cross-site") return false;
  if (!origin) return true;

  const host = request.headers.get("host");
  if (!host) return false;
  try {
    return normalizeHostname(new URL(origin).host) === normalizeHostname(host);
  } catch {
    return false;   // e.g. Origin: null
  }
}
```

The scheme carries no CSRF signal that `Host` does not already carry: a cross-origin page cannot forge the `Host` header, and `isApiRequestHostAllowed()` already constrains `Host` to loopback / IP literal / operator-configured names, which is the actual DNS-rebinding protection. Port comparison is preserved because `URL.host` includes it.

**Option B — tolerate the scheme only behind a proxy:** if `x-forwarded-proto` is present and the `Origin` host equals the `Host` header, accept either scheme.

**Option C — escape hatch:** add `PI_WEB_ALLOWED_ORIGINS` next to `PI_WEB_ALLOWED_HOSTS`. Today the allow-list only reaches the `Host` check, so operators facing a non-configurable relay have no supported option at all.

Option A or B would make pi-web work out of the box behind Azure Dev Tunnels and any other relay that normalizes `Origin` onto the backend authority.

## Related issues (not duplicates)

- **#286** — `Host` rejected behind a reverse proxy; answered with `PI_WEB_ALLOWED_HOSTS` plus "the proxy must forward `Host` and `X-Forwarded-Proto`". The Dev Tunnel relay *does* forward `X-Forwarded-Proto` but rewrites `Host` **and** `Origin`, and is not configurable, so that guidance cannot be followed. Different check (`isApiRequestHostAllowed`).
- **#329** — `Origin: null` from a sandboxed iframe, and loopback-spelling mismatch; closed via a DOCX-preview change. Option A above would have covered its spelling-mismatch case too.
- **#301** — PWA `Sec-Fetch-Site`; fixed in `d362764` for the session-export route only.
- **#492** — quick-tunnel SSE handshake regression in 0.8.8; unrelated code path (`connected` event timing), no 403.

## Current workaround

A `--require` preload (injected through `NODE_OPTIONS` in the systemd unit) patches `http.Server.prototype.emit` and rewrites `Origin` from `http://` to `https://` when `x-forwarded-proto: https` is present *and* the Origin's host already equals the `Host` header — so cross-origin requests are still rejected.

NgoQuocViet2001:
## Summary

- accept an `Origin` that differs from the request URL by scheme only, when a proxy is in front and the `Origin` authority equals the `Host` header
- keep the exact-origin comparison for direct requests

Fixes #497

## Why

A relay can report the external scheme in `x-forwarded-proto` while rewriting `Origin` onto the backend authority. Next.js derives `https` from the forwarded header, `Origin` still says `http`, and the two differ by scheme alone — so every POST is rejected with `403 Untrusted API request` even though the request is same-origin. `PI_WEB_ALLOWED_HOSTS` can't help, because the `Host` the relay sends is already loopback.

## Scope of the relaxation

The scheme is only ignored when `x-forwarded-proto` is present **and** the `Origin` authority matches `Host`. A cross-origin page cannot set `Host`, and `isApiRequestHostAllowed()` still runs first, so DNS rebinding and foreign origins are rejected exactly as before. Ports are still compared, so `localhost:30142` does not pass for `localhost:30141`.

I went with the narrower of the options in the issue (B) rather than dropping the scheme comparison outright (A), since it leaves direct-connection behaviour untouched. Happy to switch to A if you prefer it — it would also cover the loopback-spelling case from #329.

## Test plan

- Ran: `node --experimental-strip-types --test lib/request-security.test.mjs` (13 passed, 3 new)
- Ran: `npm run lint`
- Checked: with the new branch reverted, the proxy case fails and the rest still pass
- Checked: `npm test` failure count is unchanged from a clean `2a6e537` on this Windows checkout

### DIFF
--- a/lib/request-security.test.mjs
+++ b/lib/request-security.test.mjs
@@ -190,3 +190,77 @@ test("recognizes JSON request content types", async () => {
     headers: { "content-type": "text/plain" },
   })), false);
 });
+
+test("allows a proxy that reports the scheme out-of-band and rewrites Origin", async () => {
+  const { isApiRequestAllowed } = await loadSubject();
+  // Azure Dev Tunnels normalizes Host and Origin onto the backend authority
+  // but keeps the external scheme in x-forwarded-proto.
+  const request = new Request("https://localhost:30141/api/test", {
+    method: "POST",
+    headers: {
+      host: "localhost:30141",
+      origin: "http://localhost:30141",
+      "x-forwarded-proto": "https",
+      "sec-fetch-site": "same-origin",
+    },
+  });
+  assert.equal(isApiRequestAllowed(request), true);
+});
+
+test("still rejects a foreign origin when a proxy is in front", async () => {
+  const { isApiRequestAllowed } = await loadSubject();
+  const foreignHost = new Request("https://localhost:30141/api/test", {
+    method: "POST",
+    headers: {
+      host: "localhost:30141",
+      origin: "http://attacker.example",
+      "x-forwarded-proto": "https",
+      "sec-fetch-site": "same-site",
+    },
+  });
+  const foreignPort = new Request("https://localhost:30141/api/test", {
+    method: "POST",
+    headers: {
+      host: "localhost:30141",
+      origin: "http://localhost:30142",
+      "x-forwarded-proto": "https",
+      "sec-fetch-site": "same-site",
+    },
+  });
+  const alternateLoopback = new Request("https://localhost:30141/api/test", {
+    method: "POST",
+    headers: {
+      host: "localhost:30141",
+      origin: "http://127.0.0.1:30141",
+      "x-forwarded-proto": "https",
+      "sec-fetch-site": "same-site",
+    },
+  });
+  const opaque = new Request("https://localhost:30141/api/test", {
+    method: "POST",
+    headers: {
+      host: "localhost:30141",
+      origin: "null",
+      "x-forwarded-proto": "https",
+      "sec-fetch-site": "same-site",
+    },
+  });
+
+  assert.equal(isApiRequestAllowed(foreignHost), false);
+  assert.equal(isApiRequestAllowed(foreignPort), false);
+  assert.equal(isApiRequestAllowed(alternateLoopback), false);
+  assert.equal(isApiRequestAllowed(opaque), false);
+});
+
+test("does not relax the scheme without a proxy in front", async () => {
+  const { isApiRequestAllowed } = await loadSubject();
+  const request = new Request("https://localhost:30141/api/test", {
+    method: "POST",
+    headers: {
+      host: "localhost:30141",
+      origin: "http://localhost:30141",
+      "sec-fetch-site": "same-origin",
+    },
+  });
+  assert.equal(isApiRequestAllowed(request), false);
+});
--- a/lib/request-security.ts
+++ b/lib/request-security.ts
@@ -20,6 +20,20 @@ function hostnameFromAuthority(value: string): string | null {
   }
 }
 
+function normalizeAuthority(value: string): string | null {
+  if (!value || /[\s/@\\]/.test(value)) return null;
+  try {
+    const parsed = new URL(`http://${value}`);
+    if (parsed.username || parsed.password || parsed.pathname !== "/" || parsed.search || parsed.hash) {
+      return null;
+    }
+    const hostname = normalizeHostname(parsed.hostname);
+    return parsed.port ? `${hostname}:${parsed.port}` : hostname;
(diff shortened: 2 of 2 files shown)


agegr:
Superseded by #596, which preserves the original contribution and adds the same-origin Fetch Metadata guard. #596 has been merged into main.

Issue #506, opened 2026-08-15
Launcher loses the Next.js child exit signal and can silently map signal termination to exit code 0


0verme:
## What happened

The pi-web launcher exits whenever its Next.js child exits, but it discards the child exit signal and emits no diagnostic identifying why the child stopped.

In `bin/pi-web.js`:

```js
child.on("exit", (code) => process.exit(code ?? 0));
```

The listener accepts only `code`, although Node's child `exit` event also provides `signal`. If the child is terminated by a signal, `code` may be `null`; `code ?? 0` can therefore turn a signal termination into an apparent successful exit. There is also no `child.on("error")` or structured exit/close log.

This was discovered while diagnosing a Windows incident where an agent tool forcibly terminated Node processes. In that specific incident the wrapper was also killed directly, so this launcher line was not the original killer. However, the missing exit diagnostics made normal child-exit and signal hypotheses unnecessarily difficult to distinguish.

## Environment

```text
Windows 11 x64
Node.js v24.15.0
@agegr/pi-web 0.8.7
Next.js 16.2.12
```

## Expected behavior

The launcher should preserve and report both exit fields:

```js
child.on("exit", (code, signal) => {
  console.error(`[pi-web] Next.js child exited: code=${code}, signal=${signal}`);
  if (signal) {
    // Preserve a non-success result and/or re-signal appropriately.
  } else {
    process.exit(code ?? 1);
  }
});

child.on("error", (error) => {
  console.error("[pi-web] Failed to launch Next.js child:", error);
  process.exit(1);
});
```

Exact signal propagation semantics are up to the project, but a signal termination should not be silently mapped to exit code 0.

## Why this matters

Without the `(code, signal)` diagnostic, users see pi-web disappear and PowerShell return with no visible shutdown reason. This makes external termination, Next.js fatal exits, and ordinary shutdown paths hard to distinguish.

PR #515, opened 2026-08-17
fix: report why the Next.js child stopped

NgoQuocViet2001:
## Summary

- handle `child.on("error")`, which had no listener at all
- name the signal or code when the child stops on its own
- stay quiet for a shutdown we forwarded, so ctrl+c does not read as an incident

Refs #506.

## What was still open

One part of #506 is already fixed on `main`: the `code ?? 0` it quotes came from `bin/pi-web.js`, and #502 moved that into `bin/process-lifecycle.js` as `code ?? getSignalExitCode(signal)`, so a signal termination reports `128 + signum` rather than success. Worth saying explicitly, since the issue is the reason someone would look.

The other two points still stand:

**No error listener.** `wireChildProcessLifecycle` wired `exit` but never `error`. An `error` event with no listener is fatal to the wrapper itself, so a spawn failure — a bad `node`, a missing Next entry — surfaced as an unhandled-event stack trace instead of a reason. It now logs and exits 1.

**No diagnostic.** An exit the user did not ask for said nothing, which is exactly the "pi-web disappears and PowerShell returns with no visible shutdown reason" case. It now prints e.g.

```
[pi-web] Next.js exited unexpectedly (signal SIGKILL)
[pi-web] Next.js exited unexpectedly (code 1)
```

Only when we did not initiate it. Once we forward SIGINT/SIGTERM the exit is expected, so ctrl+c stays silent and a clean exit prints nothing — otherwise every ordinary shutdown would announce itself and the message would stop meaning anything.

`log` is injectable (defaults to `console.error`) so the tests assert on it rather than capturing global stderr.

## Test plan

- Ran: `node --experimental-strip-types --test lib/process-lifecycle.test.mjs` (8 passed, 4 new: spawn failure, signal exit, non-zero exit, forwarded shutdown, clean exit)
- Ran: `npm run lint`
- Checked: removing the `error` listener again fails the new spawn-failure case
- Checked: `npm test` failure count unchanged from a clean `2a6e537` on this Windows checkout

### DIFF
--- a/bin/process-lifecycle.js
+++ b/bin/process-lifecycle.js
@@ -11,11 +11,30 @@ function getSignalExitCode(signal) {
   return typeof signalNumber === "number" ? 128 + signalNumber : 1;
 }
 
-function wireChildProcessLifecycle(child, parentProcess = process, timeoutMs = shutdownTimeoutMs) {
+function describeExit(code, signal) {
+  return signal ? `signal ${signal}` : `code ${code}`;
+}
+
+function wireChildProcessLifecycle(
+  child,
+  parentProcess = process,
+  timeoutMs = shutdownTimeoutMs,
+  log = console.error,
+) {
   const signalHandlers = new Map();
   let shutdownTimer;
+  // Set once we forward a signal, so an exit we asked for stays quiet and one
+  // we did not gets reported.
+  let shuttingDown = false;
 
   const forceKill = () => child.kill("SIGKILL");
+
+  const unwire = () => {
+    if (shutdownTimer) clearTimeout(shutdownTimer);
+    for (const [forwardedSignal, handler] of signalHandlers) {
+      parentProcess.removeListener(forwardedSignal, handler);
+    }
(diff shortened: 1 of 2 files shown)


agegr:
I prepared follow-up commit agegr/pi-web@e47228b to finish the lifecycle handling:

- report every unrequested child exit, including code 0
- distinguish spawn failures from runtime ChildProcess errors
- keep the SIGKILL fallback armed when forwarding a shutdown signal fails
- remove the error listener during cleanup

Validation: node --experimental-strip-types --test lib/process-lifecycle.test.mjs (8/8 passed).

GitHub rejected a direct push because Allow edits from maintainers is disabled on this PR. Please cherry-pick e47228b, or enable maintainer edits and I can push it to fix-child-exit-diagnostics.

agegr:
Superseded by #597, which includes this change plus the child error classification and shutdown fallback fixes. #597 has been merged.

Issue #513, opened 2026-08-16
Performance: avoid full session-catalogue scans on cold list and session open


Shujakuinkuraudo:
## Problem

Pi Web is fast after the 30-second cache added in #151 is warm, but a cold or forced session-list refresh can take several seconds when the local session catalogue contains many or large JSONL histories. This delays both the sidebar and opening a saved session.

## Why opening one session is affected

On a path-cache miss, `resolveSessionPath()` and `resolveSessionIdByPath()` fall back to `listAllSessions()`. As a result, `GET /api/sessions/[id]` can enumerate and parse the whole catalogue before it opens the requested session file.

## Current behavior

- `SessionManager.listAll()` parses every session JSONL to derive list metadata such as names, message counts, and first messages.
- Pi Web then resolves project information for each distinct working directory, which can invoke Git.
- Cache expiry, explicit refresh, and lifecycle invalidation return to this expensive path.
- A cache hit is fast, while the same operation after a cache miss is visibly multi-second on a sufficiently large local history.

## Expected behavior

Opening one existing session should not wait for a full session-catalogue scan. The initial list should avoid parsing complete historical payloads on its critical path where possible.

## Suggested direction

1. Add a targeted session-id-to-path lookup that checks the expected filename/header and safely falls back to a full scan only when necessary.
2. Provide a metadata-only session discovery mode, either in the Pi SDK or Pi Web, using bounded headers and file metadata. Sidebar details such as title or first message could be enriched lazily.
3. Keep the existing cache/invalidation correctness guarantees and do not change session-file semantics.

This is related to #151, which usefully avoids repeated scans for 30 seconds, but it does not remove the cold/forced-scan cost. It is distinct from the previously fixed deep-tree serialization and long-message rendering issues.

Shujakuinkuraudo:
The targeted lookup work in #516 and #526 addresses suggestion 1 from this report: opening one known session can avoid a full catalogue scan when its storage layout is standard.

A downstream validation of suggestion 2 moved the remaining metadata cache into the Pi SDK rather than duplicating JSONL parsing in Pi Web. The process-local cache is keyed by absolute lexical file path, file size, and `mtimeMs`; it defensively snapshots mutable `Date` values, coalesces concurrent loads, does not negatively cache parse failures, and prunes entries after successful discovery. Existing sorting, progress callbacks, custom session directories, symlink aliases, and invalidation behavior remain unchanged.

On a real catalogue of 156 sessions, the SDK-level result was:

```text
[...]

PR #516, opened 2026-08-17
perf: resolve one session without scanning the catalogue

NgoQuocViet2001:
## Summary

- find a session's file from directory entries instead of parsing the catalogue
- confirm the id from the header rather than trusting the filename
- fall back to the full scan unchanged when nothing matches

This is suggestion 1 from #513, the part that makes `GET /api/sessions/[id]` stop waiting on a full scan. It does not touch `listAll()` metadata parsing (suggestion 2) or any caching/invalidation behaviour.

## Why it works

Session files are written as `<timestamp>_<id>.jsonl` under a per-project directory, so the id is already in the name. `resolveSessionPath` now lists the project directories, looks for a name ending in `_<id>.jsonl`, and reads that file's header — bounded to the first line by the existing `readSessionHeader` — to confirm `header.id` really matches. On a miss it falls through to `listAllSessions()` exactly as before, so nothing depends on the naming convention holding.

Measured against a synthetic catalogue of 320 sessions × 400 lines:

```
targeted resolve : 7 ms   (correct path)
full catalogue   : 322 ms
                   ~49x
```

## Two things I was careful about

**The name is a hint, not proof.** A file named for one id whose header says another does not resolve — the header decides. Otherwise a renamed or copied file could hand back the wrong session.

**The id never becomes a path.** It is only compared against names returned by `readdir`; nothing joins it into a path. So `../../etc/passwd` and friends cannot escape the sessions directory, which the tests pin.

## Test plan

- Ran: `node --experimental-strip-types --test lib/session-resolve.test.mjs` — 5 passed
- Ran: `npm run lint`, and `tsc --noEmit` reports nothing for `lib/session-reader.ts`
- Checked: `npm test` introduces no new failures by name against a clean `2a6e537` on this Windows checkout
- Checked: the new suite fails without the change — the targeted case falls back and the assertion on the fast path does not hold

Cases cover a hit, a hit in a different project directory, the header/name mismatch above, an unknown id returning `null`, and the traversal attempts.

### DIFF
--- a/lib/session-reader.ts
+++ b/lib/session-reader.ts
@@ -4,8 +4,9 @@ import {
   buildSessionContext as piBuildSessionContext,
   getAgentDir,
 } from "@earendil-works/pi-coding-agent";
-import { closeSync, openSync, readSync } from "fs";
-import { normalize as normalizePath } from "path";
+import { closeSync, existsSync, openSync, readSync } from "fs";
+import { readdir } from "fs/promises";
+import { join as joinPath, normalize as normalizePath } from "path";
 import type { AgentMessage, SessionEntry, SessionHeader, SessionInfo, SessionContext } from "./types";
 import type { SessionEntry as PiSessionEntry, SessionInfo as PiSessionInfo } from "@earendil-works/pi-coding-agent";
 import { normalizeToolCalls } from "./normalize";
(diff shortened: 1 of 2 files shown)


agegr:
Thanks for the contribution. We’ll close this PR in favor of #526, which covers both forward and reverse session resolution and handles ambiguous paths.

PR #531, opened 2026-08-17
feat(cli): support Pi agent startup options

sadkf:
## Summary

Pi Web currently accepts only server flags (`--port`, `--hostname`, and `--no-open`) and silently ignores Pi flags such as `-ns` and `-nc`. This makes it impossible to start Pi Web with the same resource, model, and tool constraints used by the Pi CLI.

This PR:

- adds Pi-compatible Agent startup flags for models, thinking, tools, extensions, skills, prompt templates, Agent themes, context files, and system prompts
- supports Pi's multi-character aliases, including `-ns`, `-nc`, `-nt`, `-nbt`, and `-xt`
- forwards validated options from the launcher to every in-process `AgentSession` runtime
- keeps the model selector and Agent runtime on the same model/resource configuration
- resolves relative resource paths from the directory where `pi-web` was launched
- treats launcher tool restrictions as server policy, so browser presets cannot bypass them
- rejects unknown flags instead of silently ignoring them and adds `pi-web --help`
- keeps TUI/session-only flags explicitly unsupported because Pi Web manages multiple persistent sessions itself
- strips internal `PI_WEB_*` values (including the option handoff and Web password) from Agent bash subprocesses
- documents the new behavior in English and Simplified Chinese

Example:

```bash
# Do not load skills or AGENTS.md / CLAUDE.md into Agent prompts
pi-web -ns -nc

# Read-only tools with a constrained model scope
pi-web \
  --models 'anthropic/*:high,openai/gpt-*' \
  --tools read,grep,find,ls \
  --exclude-tools bash
```

## Supported Agent flags

- Model: `--provider`, `--model`, `--thinking`, `--models`, `--api-key`
- Tools: `--tools/-t`, `--exclude-tools/-xt`, `--no-builtin-tools/-nbt`, `--no-tools/-nt`
- Resources: `--extension/-e`, `--no-extensions`, `--skill`, `--no-skills/-ns`, `--prompt-template`, `--no-prompt-templates`, `--theme`, `--no-themes`, `--no-context-files/-nc`
- Prompt: `--system-prompt`, `--append-system-prompt`
- Runtime: `--offline`, `--help/-h`

`--print`, `--continue`, `--resume`, `--session`, `--fork`, and `--no-session` remain unsupported because their single-process/TUI semantics do not map cleanly to Pi Web.

## Validation

- `npm test` — 598 tests passed
- `npm run lint`
- `git diff --check`

A dedicated SDK integration test verifies that `noSkills` and `noContextFiles` reach Pi's `DefaultResourceLoader` and produce no loaded skills or context files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,52 @@ For example:
 pi-web -p 8080 -H 0.0.0.0 --no-open
 ```
 
+### Agent startup options
+
+Pi Web also accepts the Pi options that configure `AgentSession` creation. They
+act as server-wide defaults whenever Pi Web opens an Agent runtime. An explicit
+model or thinking level selected in the browser for a new session takes
+precedence over the corresponding startup default. Launcher tool allowlists and
+suppression flags are server policy; browser presets cannot re-enable tools they
+disable, and `--exclude-tools` is always applied as a denylist.
+
+| Category | Options |
+| --- | --- |
+| Model | `--provider <name>`, `--model <pattern>`, `--thinking <level>`, `--models <patterns>`, `--api-key <key>` |
+| Tools | `--tools <names>` / `-t`, `--exclude-tools <names>` / `-xt`, `--no-builtin-tools` / `-nbt`, `--no-tools` / `-nt` |
+| Extensions | `--extension <source>` / `-e` (repeatable), `--no-extensions` |
+| Skills | `--skill <path>` (repeatable), `--no-skills` / `-ns` |
+| Prompt templates | `--prompt-template <path>` (repeatable), `--no-prompt-templates` |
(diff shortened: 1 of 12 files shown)


agegr:
I don’t think Pi Web needs full parity with Pi’s CLI startup options. Pi Web is a long-running, multi-session server, while most of these flags are designed for a single CLI session and already have equivalents in the Web UI or settings.

It may be better to scope this PR down to server-level restrictions that cannot be configured elsewhere, such as tool policies, `--no-skills`, and `--no-context-files`. This would reduce duplicated configuration and avoid semantic drift from the upstream CLI.

PR #533, opened 2026-08-17
fix: avoid duplicate full-size image render in user messages

epheien:
## Problem

When a user attaches an image, pi writes the attached image's path back into the user message text as a Markdown image reference (e.g. ``). With image blocks present, that reference was rendered **a second time at full size** through `MarkdownBody`'s local-path resolver (`/api/files`), clipping inside the bubble's max-height — the same picture appeared twice in one user message:

1. the image-block thumbnail (clickable preview), and
2. a cropped full-size duplicate below/above the text.

## Fix

`UserMessageView` now strips Markdown image references from the text when image blocks are present (the picture is already rendered from the blocks). When there are no image blocks, local Markdown images keep working as before.

## Tests

- `does not re-render the markdown image reference when image blocks exist` — fails on the old code (asserts no second `/api/files` image), passes with the fix; red-green verified against the parent commit.
- `still renders local markdown images when there are no image blocks` — regression guard for the non-image-block path.
- Full `MessageView.test.mjs` suite: 10/10 pass; `tsc --noEmit` and `eslint` clean.

## Known trade-off

When image blocks exist, *all* Markdown image references in the text are stripped, including ones manually written by the user (a rare case: hand-written local image reference + attached image in the same message). Since base64 image blocks don't carry the original file path, a precise "only strip the injected reference" match isn't reliably possible. Happy to refine if reviewers consider it necessary.

### DIFF
--- a/components/MessageView.test.mjs
+++ b/components/MessageView.test.mjs
@@ -3,6 +3,11 @@ import test from "node:test";
 import React from "react";
 import { renderToStaticMarkup } from "react-dom/server";
 import { createJiti } from "jiti";
+import path from "node:path";
+
+// Project root, derived from this file's location so tests never hardcode
+// machine-specific absolute paths.
+const PROJECT_ROOT = path.resolve(import.meta.dirname, "..");
 
 const jiti = createJiti(import.meta.url, {
   jsx: { runtime: "automatic" },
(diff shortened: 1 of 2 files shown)


epheien:
## Reproduction (before the fix)

The same picture appears twice in one user message: once as the image-block thumbnail, once as a full-size image rendered from the Markdown reference pi injects into the text — cropped by the bubble's `max-height`:

Duplicate image render before fix

After this fix, only the thumbnail (clickable for a larger preview) is rendered.

agegr:
I’m unable to reproduce the premise that Pi automatically writes an attached image path back into the user message as Markdown.
In the current pi-web flow, the original message text is preserved and images are sent separately via { message, images }. I also couldn’t find this Markdown injection in the locked @earendil-works/pi-coding-agent@0.84.2; its file processor uses <file name="...">.
In an attempted reproduction, the persisted user message contained only the manually written Markdown reference. There was no user image block; the later image block belonged to a read tool result. Supplying both an attachment and a Markdown reference manually is two image inputs, so rendering both is expected.
Could you provide exact reproduction steps or a minimal real JSONL entry where attaching an image once produces both:
[
  { "type": "text", "text": "... " },
  { "type": "image", "data": "...", "mimeType": "image/png" }
]
Without evidence of that producer path, this fix changes valid behavior and may remove manually authored Markdown images, image syntax inside code blocks, and parts of paths containing parentheses.

PR #543, opened 2026-08-18
feat(sidebar): persist user-pinned custom project paths in localStorage

3bluebytes:
## Problem

The directory picker's "Custom path" branch validates a path server-side but only stores the choice in component state. The path vanishes on reload and is absent from the project selector, so users have to re-pick the same directory every time they open Pi Web.

Project list (`getRecentProjects`) is computed purely from session files on disk, so any path the user has only ever browsed (no session started) never appears in the selector.

## Fix

Pin successfully-validated custom paths in `window.localStorage["pi-web:custom-project-paths"]` and merge them into `getRecentProjects`. Session-derived projects still win on overlap because they are canonicalised against the server's filesystem view.

- New `lib/custom-projects.ts` with `loadCustomProjects` / `saveCustomProject` / `removeCustomProject` helpers. Falls back gracefully when storage is unavailable (privacy mode, SSR, denied access).
- `getRecentProjects` now accepts an optional `customProjects` list and exposes `isCustomOnlyProject` so the sidebar can show a remove button only for entries not backed by any session.
- The sidebar calls `saveCustomProject` after a successful `/api/cwd/validate` and refreshes state from storage. A small × button removes a pinned entry on hover.
- i18n: new `sidebar.removeCustomProject` key in en + zh-CN.

## Tests

- New `lib/custom-projects.test.mjs` covers round-trip, dedup-by-key, malformed entries, removal, and storage-unavailable / SSR fallback.
- New cases in `lib/project-groups.test.mjs` cover the merge logic and `isCustomOnlyProject`.
- Full suite (`npm test`): **597/597 pass**.
- `npm run lint`: clean.
- `npm run build`: clean (no new warnings).

## Screenshots

Will attach after first reply.

### DIFF
--- a/components/SessionSidebar.tsx
+++ b/components/SessionSidebar.tsx
@@ -5,7 +5,8 @@ import type { SessionInfo } from "@/lib/types";
 import { loadExplorerOpen, saveExplorerOpen } from "@/lib/file-explorer-state";
 import { dispatchSessionRowContextMenu } from "@/lib/session-row-context-menu";
 import { skillExpansionToCommand } from "@/lib/slash-display";
-import { getProjectActivity, getRecentProjects, sessionsForProject } from "@/lib/project-groups";
+import { getProjectActivity, getRecentProjects, isCustomOnlyProject, sessionsForProject } from "@/lib/project-groups";
+import { loadCustomProjects, removeCustomProject, saveCustomProject } from "@/lib/custom-projects";
 import { workspaceKeyOf } from "@/lib/workspace-memory";
 import { useI18n } from "@/hooks/useI18n";
 import { DirectoryPicker } from "./DirectoryPicker";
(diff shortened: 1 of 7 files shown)


agegr:
I don’t think this is necessary. The project list should only include directories with existing sessions; custom paths can remain temporary until a session is created.

agegr:
A simpler alternative would be to remember the last successfully selected cwd and prefill the custom picker with it. This avoids adding directories without sessions to the project list.

a5630c15f9c2df8cce6af1c996e53dd29084e9fa

3bluebytes:
> A simpler alternative would be to remember the last successfully selected cwd and prefill the custom picker with it. This avoids adding directories without sessions to the project list.
> 
> [a5630c1](https://github.com/agegr/pi-web/commit/a5630c15f9c2df8cce6af1c996e53dd29084e9fa)

nice method, thanks for fixing

PR #544, opened 2026-08-18
fix(security): tolerate Chromium stripping port from Origin header

### LINKED ISSUE #542: Bug: All API requests return 403 on Chromium 150+ browsers (PR #496 has been submitted)
### Environment
- pi-web version: 0.8.9 (commit `2a6e537`)
- Node.js: ≥ 22.19.0
- Browser: Chrome 150+ / Edge 150+ / other Chromium 150+
- URL: `http://127.0.0.1:30141` (default port)

### Description
After loading the UI, every `/api/*` request is rejected with
`403 Untrusted API request`. The session sidebar is empty,
`Error: HTTP 403` is displayed, and the model config panel cannot load.

### Cause
`lib/request-security.ts:isApiRequestOriginAllowed` uses
`canonicalOrigin(origin) === requestOrigin` (strict, includes port).

Chromium 150+ strips the port from the `Origin` header for same-origin
requests on non-default ports. Since pi-web defaults to port 30141,
every request now fails the comparison.

Verified headers from a stock install:

    Host: 127.0.0.1:30141
    origin: http://127.0.0.1
    Sec-Fetch-Site: same-origin

### Suggested Fix
Compare Origin hostnames instead of full canonical origins (the Host
header is already separately validated). I have a working patch +
tests locally and am happy to open a PR.

LCZcoding:
## Summary

`isApiRequestOriginAllowed` (`lib/request-security.ts`) rejects all `/api/*`
requests with 403 on Chromium 150+, because Chrome now strips the port from
the `Origin` header for same-origin requests on non-default ports, and the
canonical-origin comparison (which includes the port) no longer matches.

pi-web defaults to port 30141, so every fresh install using a modern
Chromium-based browser is affected: the UI shows `Error: HTTP 403`, the
session sidebar appears empty, and the model config panel cannot load.

Fixes #542 

## Repro

1. `npm install && npm run dev` on a fresh checkout
2. Open `http://127.0.0.1:30141` in Chrome 150+
3. DevTools → Network → any `/api/*` request → status 403,
   response `{"error":"Untrusted API request"}`
4. Request headers show `Origin: http://127.0.0.1` (no port) vs
   `Host: 127.0.0.1:30141`

## Fix

Compare Origin hostnames instead of full canonical origins. The `Host`
header is already validated by `isApiRequestHostAllowed`, so port
differences carry no trust signal — only hostname equality does.

Add `originHostname()` helper and switch `isApiRequestOriginAllowed`
to use hostname comparison.

## Security

The Host allowlist, `sec-fetch-site`, and same-origin checks still
block DNS rebinding, cross-site fetches, opaque iframes, and cross
loopback-name origins (covered by new regression tests). The only
loosening is accepting same-hostname / different-port Origin headers,
which is the intended Chromium behavior.

## Tests

- New: `allows same-origin requests when Chromium strips the port from Origin`
  (covers loopback IP, LAN IP, loopback name)
- New: `still rejects when Origin hostname differs from Host even with port stripped`
  (covers cross-loopback-name and DNS rebind)

Existing tests continue to pass.

### DIFF
--- a/lib/request-security.test.mjs
+++ b/lib/request-security.test.mjs
@@ -178,6 +178,70 @@ test("rejects missing, malformed, and unconfigured Host headers", async () => {
   })), false);
 });
 
+test("allows same-origin requests when Chromium strips the port from Origin", async () => {
+  const { isApiRequestAllowed } = await loadSubject();
+  // Loopback IP with the port stripped from Origin (Chromium 150+ behavior).
+  const loopback = new Request("http://127.0.0.1:30141/api/agent/running", {
+    headers: {
+      host: "127.0.0.1:30141",
+      origin: "http://127.0.0.1",
+      "sec-fetch-site": "same-origin",
+    },
+  });
+  assert.equal(isApiRequestAllowed(loopback), true);
+
+  // LAN IP with the port stripped from Origin.
+  const lan = new Request("http://192.168.32.7:30141/api/test", {
+    method: "POST",
+    headers: {
+      host: "192.168.32.7:30141",
+      origin: "http://192.168.32.7",
+      "sec-fetch-site": "same-origin",
+      "content-type": "application/json",
+    },
+  });
+  assert.equal(isApiRequestAllowed(lan), true);
+
+  // Loopback name with the port stripped from Origin.
+  const named = new Request("http://localhost:30141/api/test", {
+    method: "POST",
+    headers: {
+      host: "localhost:30141",
+      origin: "http://localhost",
+      "sec-fetch-site": "same-origin",
+      "content-type": "application/json",
+    },
+  });
+  assert.equal(isApiRequestAllowed(named), true);
+});
+
+test("still rejects when Origin hostname differs from Host even with port stripped", async () => {
+  const { isApiRequestAllowed } = await loadSubject();
+  // Cross-loopback-name origin should not slip through just because both
+  // sides happen to be loopback addresses.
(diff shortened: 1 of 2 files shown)


agegr:
I couldn’t reproduce this. Chrome 151 includes the non-default port in the `Origin` header for same-origin POST requests. Same-origin GET/HEAD requests normally omit `Origin` entirely.

Could you provide a HAR or raw headers from the failing POST request, including the exact Chrome version? Comparing hostname only would allow cross-port same-site requests and weaken the CSRF protection.

LCZcoding:
Superseded by #615, which addresses the cross-port CSRF concern you raised.

Two pieces of evidence worth noting:

1. The original reproduction was on Chrome **152.0.0.0** (`sec-ch-ua: "Google Chrome";v="152"`), not 151. Chrome 152 strips the port on GET *and* POST alike — confirmed by a `POST /api/default-cwd` capture pairing `Origin: http://127.0.0.1` with `Host: 127.0.0.1:30141`. The `Referer` header on the same request also drops the port.

2. `gofixpoint/amika#373` independently described this as "a separate upstream bug" against #542, so the behavior is reproducible beyond this PR.

The new patch keeps the port check strict when Origin carries a port, so cross-port CSRF (`Origin: http://myapp.com:8080` vs `Host: myapp.com:30141`) is still rejected. It also builds on the proxy-rewrite handler that landed on main, so the two paths compose cleanly.

PR #548, opened 2026-08-19
feat: add maximized file panel layout

looluo:
## Summary

- add a maximize/restore control to the open file panel header, immediately left of the panel visibility control
- let the file panel replace the conversation area while keeping the conversation mounted so active agent runs continue
- preserve sidebar controls, stored panel width, responsive behavior, blocking-attention recovery, and keyboard focus
- add English/Chinese labels and focused regression coverage

## Testing

- `npm run lint`
- `node --test components/AppShell.file-panel-maximize.test.mjs components/AppShell.file-viewer-state.test.mjs components/AppShell.mobile-toolbar.test.mjs components/MobilePwaLayout.test.mjs` (18 passed)
- browser QA at 1264px, 800px, and 600px, including maximize/restore focus transfer and responsive layout transitions

## Full-suite note

`npm test` was run, but the current baseline has unrelated failures in component render tests that call `useI18n` without an `I18nProvider`, plus a Git Bash PATH-separator assertion in `lib/project-command-env.test.mjs`.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -1245,6 +1245,12 @@ span.token.table {
 .right-panel-overlay-backdrop {
   display: none;
 }
+.conversation-panel {
+  display: flex;
+}
+.conversation-panel.file-panel-maximized {
+  display: none;
+}
 
 /* Desktop sidebar: animate width */
 @media (min-width: 641px) {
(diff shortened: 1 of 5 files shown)


agegr:
Thanks for the contribution. After testing, the maximize control doesn’t add much value because the file panel can already be resized. It also has little visible effect, or can even make the panel narrower, at some compact desktop widths. I’m going to close this PR for now.

PR #550, opened 2026-08-19
chore: tighten TypeScript guards, silence stray rejections

haotianliangye:
A grab bag of small changes that keep the dev experience clean under the strict TypeScript / Next 16 dev overlay:

TypeScript type guards on `AssistantMessage.content`:
- app/api/models-config/test/route.ts
- components/ChatWindow.tsx
- lib/session-title.ts

`@earendil-works/pi-ai`'s `AssistantMessage.content` is typed as `AssistantContentBlock[]`, which is a discriminated union. The naive `.filter((b) => b.type === "text")` produced a `(AssistantContentBlock | { type: 'text'; text: string })[]` that the compiler still treated as the wide type. Switching to `.filter((b): b is TextContent => ...)` narrows the result so the downstream `.map((b) => b.text)` typechecks without an `as TextContent` cast.

hooks/useTheme.ts:
- `document.startViewTransition` is not yet in the default lib types on every platform we ship to. Replace the unguarded access with a type guard, and bail to the synchronous `apply()` path when the API is missing so the theme still flips on browsers without View Transitions.
- No runtime behavior change for browsers that have the API.

hooks/useAgentSession.ts:
- `loadModels` already swallows its own errors (network failures, JSON parse failures, AbortError on cleanup). Add a defensive `.catch(() => {})` so anything that ever bubbles up is silenced — Next 16's dev overlay still reports the AbortError source line even when caught, which previously surfaced as an unhandledRejection noise.

components/ChatWindow.tsx:
- Same AbortError-catching comment near the new-session update check.
- Same TextContent guard on the user-message text extraction path.

### DIFF
--- a/app/api/models-config/test/route.ts
+++ b/app/api/models-config/test/route.ts
@@ -3,6 +3,7 @@ import { mkdtempSync, rmSync, writeFileSync } from "fs";
 import { tmpdir } from "os";
 import { join } from "path";
 import { completeSimple, type AssistantMessage } from "@earendil-works/pi-ai/compat";
+import type { TextContent } from "@earendil-works/pi-ai";
 import { ModelRuntime } from "@earendil-works/pi-coding-agent";
 import { hasJsonContentType, isApiRequestAllowed } from "@/lib/request-security";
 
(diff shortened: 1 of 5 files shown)


agegr:
Thanks for the contribution. I checked this against the current toolchain: the base already passes TypeScript checks, the existing View Transition guard handles unsupported browsers, and the Promise catch has the same runtime behavior. Since this does not provide a concrete fix, we will close the PR for now.

PR #551, opened 2026-08-19
feat: PWA support + QR pairing + Tailscale auto-discovery for mobile access

haotianliangye:
# feat: PWA support + QR pairing + Tailscale auto-discovery for mobile access

## Summary

Make Pi Web a true mobile-first interface. Adds a Progressive Web App shell so a phone can install Pi Web to its home screen, replaces manual password typing with a one-time QR pairing token, and has the desktop launcher auto-discover its Tailscale IP so the QR code points at the right address without any user configuration.

This PR combines three commits that evolved together in a downstream fork:

1. **`feat: connect phone via QR + Tailscale`** — runtime password, pairing tokens, Connect Phone modal, Tailscale-aware launcher
2. **`feat(pwa): install prompt, iOS hint, update toast`** — user-facing UI for installing the PWA and detecting updates
3. **`feat(pwa): manifest, icons, screenshots, SW cache hygiene`** — PWA infrastructure (manifest, icons, service worker, offline page)

---

### The current mobile flow is fragile

Today, the only documented way to reach Pi Web from a phone is:

```bash
PI_WEB_PASSWORD='your-password-here' pi-web --hostname 0.0.0.0
# Then on the phone, type the LAN IP, type the password
```

Three things go wrong:

- **"What's my LAN IP?"** — users guess `192.168.1.x`, get it wrong, give up.
- **"What's the password?"** — multi-tap typing on a phone keyboard for a complex password is a chore.
- **Tailscale users have it worse** — the launcher binds to `0.0.0.0` but the printed URL is still `127.0.0.1`. If they bind to `100.x.x.x` they have to remember which one.

### The QR code path fixes all three

- **No more guessing IPs** — the QR encodes the right address.
- **No more password typing on the phone** — the QR contains a one-time pairing token that exchanges for a 30-day session cookie.
- **Tailscale auto-resolved** — the launcher runs `tailscale ip -4` (or falls back to the first non-loopback IPv4) and uses *that* in the QR code.

### Why a Progressive Web App and not a native app

- A PWA installs from the browser with one tap — no app store review cycle, no platform-specific binaries.
- The same `pi-web` binary is still the runtime; the PWA is just an additional install surface.
- A native app would have to keep up with four platform-specific codebases (iOS/Android/macOS/Windows) and would re-introduce the desktop-client problem this whole project tries to avoid.

---

#### Files

| File | Purpose |
|---|---|
| `app/manifest.ts` | Web App Manifest: name, icons, screenshots, `display: "standalone"`, `start_url`, `scope`, shortcuts |
| `public/icons/icon-192.png` | PWA icon (any size) |
| `public/icons/icon-512.png` | PWA icon (high-res) |
| `public/icons/icon-512-maskable.png` | Maskable variant for Android adaptive icons |
| `public/icons/shortcut-new.png` | Icon used by the "New session" home-screen shortcut |
| `public/screenshots/desktop-wide.png` | Wide screenshot referenced in `screenshots[]` |
| `public/screenshots/mobile-narrow.png` | Narrow screenshot for mobile install banners |
| `public/sw.js` | Service worker — precache + cache-first for static, never for `/api/*` |
| `public/sw.test.mjs` | Tests for `sw.js` notification-click handler and message handler |
| `public/offline.html` | Fallback page when the user goes offline mid-session |

#### Manifest contract (`app/manifest.ts`)

```ts
{
  id: "/",
  name: "Pi Web",
  short_name: "Pi Web",
  start_url: "/",
  scope: "/",
  display: "standalone",            // ← key: no browser chrome after install
  orientation: "any",
  background_color: "#1a1a1a",
  theme_color: "#1a1a1a",
  icons: [
    { src: "/icons/icon-192.png", sizes: "192x192", type: "image/png", purpose: "any" },
    { src: "/icons/icon-512.png", sizes: "512x512", type: "image/png", purpose: "any" },
    { src: "/icons/icon-512-maskable.png", sizes: "512x512", type: "image/png", purpose: "maskable" }
  ],
  screenshots: [...],
  shortcuts: [{ name: "New session", url: "/?action=new", ... }]
}
```

#### Service worker behaviour (`public/sw.js`)

- **Install**: clears all `pi-web-*` caches (handles pnpm→npm migration residue) and precaches `PRECACHE_URLS` (manifest + icons + offline page).
- **Activate**: deletes old `pi-web-*` caches, calls `clients.claim()`.
- **Fetch**:
  - **Always bypasses** `/api/*` and `/sw.js` itself.
  - **Navigate requests** — try network, fall back to `offline.html`.
  - **Static assets** (`/_next/static/*` or anything in `PRECACHE_URLS`) — cache-first.
- **notificationclick** — focuses an existing client at the notification's URL, or opens a new window. Cross-origin targets are rejected.
- **message** — handles `{ type: "SKIP_WAITING" }` from the registration helper.

### 2. PWA user-facing UI (`feat(pwa): install prompt, iOS hint, update toast`)

| File | Purpose |
|---|---|
| `components/PwaRegistration.tsx` | Registers the SW on mount. In dev mode, unregisters any stale SW from a previous `npm start` and reloads — cached production chunks otherwise break dev. |
| `components/PwaInstallPrompt.tsx` | Captures `beforeinstallprompt`, defers it, surfaces an "Install Pi Web" chip in the corner until the user clicks (or `appinstalled` fires). |
| `components/PwaIosHint.tsx` | iOS Safari doesn't fire `beforeinstallprompt`. Detects iOS Safari and shows "Share → Add to Home Screen" instead. |
| `components/PwaUpdateToast.tsx` | Listens for `controllerchange` and shows a "New version available — reload" toast. |
| `components/PwaManifest.test.mjs` | Tests for manifest shape |

The install chip and iOS hint are deliberately minimal — small chip in the corner, not a modal banner — so they don't interfere with the chat UI.

#### Runtime password (`lib/runtime-password.ts`)

Why a *file* and not `globalThis`:

> Next.js runs the proxy middleware in a *separate worker process* from the route handlers. `globalThis` is not shared across processes. The smallest cross-process bridge is a file.

Format: 6 decimal digits, mode 0o600, written atomically (`writeFileSync` to a temp file then `rename`). Cold boot wipes the file (intentional — invalidates prior session cookies, this is the "switch device and re-auth" path).

Resolution order: `process.env.PI_WEB_PASSWORD` (opt-in override) → existing file → generate + persist + return.

Note on no in-process cache: even within one process, Next.js 16 may compile `proxy.ts` and route files into separate module graphs that don't share `globalThis`. A cached password would diverge between them and Basic Auth would silently fail. The file is the only cross-process source of truth. A 6-byte read from local disk is cheap enough that we just always read.

#### Pairing token flow

```
desktop                              phone
  │                                    │
  ├─ POST /api/pair-info (operator)    │
  │   → { address, pinRequired,        │
  │       pairingToken, expiresIn:30s }│
  │                                    │
  │   ┌─── QR encodes:                 │
  │   │   <address>/?pair=<token>      │
  │   └─── operator scans with phone ───┤
  │                                    │
  │   ┌─── phone browser opens URL ────┤
  │   │   GET /?pair=<token>           │
  │   │   → handler validates token    │
  │   │   → sets 30-day cookie         │
  │   │   → redirects to /             │
  │   └─── session established ────────┤
  │                                    │
  │   (PIN still works as fallback:    │
  │    if phone lands via bare URL,    │
  │    the Connect Phone modal shows   │
  │    the 6-digit PIN; user types it) │
```

#### New API routes

| Route | Method | Purpose |
|---|---|---|
| `/api/pair-info` | GET | Returns `{ address, pinRequired, pairingToken, expiresIn }`. `address` is the Tailnet IP if Tailscale is up, else the first non-loopback IPv4, else `127.0.0.1`. |
| `/api/pair-password` | GET | Returns `{ present, source, regeneratable }` by default. With `?reveal=1` AND `isApiRequestAllowed(req)` passing, returns the actual PIN. The `isApiRequestAllowed` gate blocks passive exfiltration (rogue extensions, malicious pages). |
| `/api/pair-password/regenerate` | POST | Operator-triggered PIN rotation. Invalidates existing session cookies. The "switch device and re-auth" path. |
| `/api/pair-tokens` | GET (or POST) | Exchanges a one-time `pair` query token for a `pi-session` cookie (30-day expiry). |

#### `bin/pi-web.js` changes

- On startup with `-H 0.0.0.0`: run `tailscale ip -4`. If it succeeds, use the returned IP for the QR code. If `tailscale` is not installed, fall back to the first non-loopback IPv4. If neither works, fall back to `127.0.0.1` (with a hint that the phone won't reach it).
- On cold boot: delete the runtime password file. Same rationale — invalidate prior cookies.
- Keep the existing `openBrowserWindow` behaviour (auto-open the desktop browser when "Ready" appears in stdout).

#### Connect Phone modal (`components/PairDevice.tsx`)

- Top-bar icon → opens the modal.
- Modal shows the QR code (encodes `<address>/?pair=<token>`) plus the 6-digit PIN as a fallback.
- After successful pairing, the desktop session lives in a 30-day cookie.
- "Regenerate" button rotates the PIN and invalidates existing cookies.

---

## Test plan

> All commands below assume `npm install && npm run dev:lan` (or `npm run build && npm run start:lan` for production-behaviour tests).

### A. PWA install on Chrome Android (desktop view)

1. Open `http://<lan-ip>:30141/` on the desktop.
2. After a few seconds, an "Install Pi Web" chip appears in the bottom corner.
3. Click it → Chrome shows the install prompt → accept → app installs to the launcher.
4. Open from the launcher → app opens standalone (no address bar, no tab UI).

### B. PWA install on iOS Safari

1. Open `http://<lan-ip>:30141/` on the phone in Safari.
2. The "iOS hint" chip appears in the corner (instead of the install chip).
4. Tap Share → Add to Home Screen → app installs.
5. Open from the home screen → app opens standalone.

### C. QR pairing with Tailscale

1. On the desktop, ensure Tailscale is running and you have a tailnet address (e.g. `100.64.1.2`).
2. Start the server bound to the tailnet IP:
   ```bash
   pi-web --hostname 100.x.x.x        # the tailnet IP
   ```
3. Click the "Connect Phone" icon in the top bar.
4. The modal shows:
   - A QR code encoding `https://100.x.x.x:30141/?pair=<token>`
   - The 6-digit PIN as a fallback
5. On the phone, open the Camera app, point at the QR.
6. Phone lands directly on the desktop session — no password typed.

### D. QR pairing on LAN (no Tailscale)

1. Same flow as C, but `tailscale` is not installed:
   ```bash
   pi-web --hostname 0.0.0.0
   ```
2. The launcher runs `tailscale ip -4`, gets a non-zero exit code, and falls back to the first non-loopback IPv4 (e.g. `192.168.1.42`).
3. QR encodes that LAN IP.

### E. Bare URL fallback (PIN entry)

1. With Tailscale up, start the server bound to the tailnet IP.
2. On the phone, *manually* open the URL `http://100.x.x.x:30141/` (don't use the QR).
3. The Connect Phone modal opens on the phone, prompting for the 6-digit PIN.
4. Type the PIN displayed on the desktop → phone session authenticated, cookie set for 30 days.

### F. Regeneration

1. On the desktop, click "Regenerate" in the Pair Device modal.
2. The PIN rotates.
3. Any phone with a 30-day cookie gets `401` on the next request.
4. Phone re-prompts for the new PIN.

### G. PWA update toast

1. While the dev server is running, edit any file under `app/` → save.
2. (For production: deploy a new build.)
3. The service worker installs the new version → "New version available — reload" toast appears on the desktop.
4. Click reload → app loads the new version.

### H. Dev mode cache hygiene

1. Start `npm run start` (production server). SW registers.
2. Stop it, start `npm run dev`.
3. First GET `/` triggers `PwaRegistration`'s dev-mode cleanup: unregisters the production SW, reloads the page.
4. After reload, `npm run dev` works normally (cached production chunks don't break dev).

### I. Existing flow still works

- `PI_WEB_PASSWORD='your-password' pi-web --hostname 0.0.0.0` → unchanged behaviour, Basic Auth still applies.

---

### PIN is reachable only via `?reveal=1` + same-origin

`isApiRequestAllowed(req)` (in `lib/request-security.ts`) blocks passive fetches:

- A malicious page embedding `` — blocked (cross-origin).
- A rogue browser extension calling `fetch('/api/pair-password?reveal=1')` — blocked (no `Sec-Fetch-Site: same-origin`).
- The in-app Connect Phone modal — allowed (same-origin, top-level navigation).

### Pairing tokens are short-lived (~30s)

The QR contains `<address>/?pair=<token>`. The token is single-use and expires fast. If the QR is photographed by a third party, the token is useless within seconds.

### Brute-force risk on the 6-digit PIN

A 6-digit decimal PIN has 10⁶ = 1M combinations. Mitigations:

- The server is reached via Tailscale (or LAN), which is its own authentication layer — the network is closed to the public internet by default.
- Each wrong attempt is rate-limited (existing Basic Auth rate limiting applies).
- The PIN is auto-rotated by `regenerate` on demand.

This is acceptable for a developer tool accessed on a trusted tailnet. It is **not** acceptable for a public-internet deployment — the existing HTTPS-via-reverse-proxy guidance still applies.

### Session cookie

`pi-session` cookie is HttpOnly, 30-day expiry, scoped to the origin. Cleared by regenerate.

### What this PR does NOT change

- TLS / Basic Auth behaviour for `PI_WEB_PASSWORD` users.
- `/api/*` authentication beyond the PIN layer.
- Proxy middleware behaviour for non-mobile clients.

---

## Breaking changes

For users who **bind to a non-loopback address without setting `PI_WEB_PASSWORD`**:

- **Before**: server was reachable with no auth at all.
- **After**: server requires the PIN (displayed in the Connect Phone modal) or a valid `pi-session` cookie.

This is a security improvement, but it is a default-behaviour change. It should be called out in the release notes. For users who hit this, the migration path is:

- Open the desktop UI → top bar → "Connect Phone" → modal shows the PIN.
- On the phone, either scan the QR or type the PIN.

For users who **set `PI_WEB_PASSWORD`**, behaviour is unchanged.

---

## Files

**41 files changed, ~2178 lines added**.

### New files (23)

```
app/api/pair-info/route.ts
app/api/pair-password/route.ts
app/api/pair-password/regenerate/route.ts
app/api/pair-tokens/route.ts
app/manifest.ts (already exists upstream; PR modifies the manifest shape)
components/PairDevice.tsx
components/PwaInstallPrompt.tsx
components/PwaInstallPrompt.test.mjs
components/PwaIosHint.tsx
components/PwaIosHint.test.mjs
components/PwaManifest.test.mjs
components/PwaRegistration.tsx
components/PwaUpdateToast.tsx
components/PwaUpdateToast.test.mjs
lib/pair-tokens.ts
lib/runtime-password.ts
lib/session-store.ts
lib/session-store.test.mjs
public/icons/icon-192.png
public/icons/icon-512.png
public/icons/icon-512-maskable.png
public/icons/shortcut-new.png
public/offline.html
public/screenshots/desktop-wide.png
public/screenshots/mobile-narrow.png
public/sw.js (already exists upstream; PR replaces with the full version)
public/sw.test.mjs
```

### Modified files (18)

```
.gitignore
README.md
README.zh-CN.md
README.ja.md
README.ru.md
app/globals.css
app/layout.tsx
bin/pi-web-options.js
bin/pi-web.js
components/AppShell.tsx
lib/i18n/messages/en.ts
lib/i18n/messages/zh-CN.ts
lib/web-auth.ts
next.config.ts
package.json
proxy.ts
```

---

## Out of scope (deliberately left for follow-up PRs)

- **Push notifications for completed sessions.** The SW supports `notificationclick` but the SW doesn't push proactively. Web Push requires a server-side VAPID setup that should be a separate PR.
- **Native share-target integration.** The manifest's `share_target` field is not set; receiving shared text/images into Pi Web is a separate feature.
- **Tailscale ACL scoping per session.** Currently the server trusts the Tailscale layer entirely. Per-session ACLs (e.g. "this session can only read files in `/foo`") is unrelated to mobile access.
- **App-store-grade icon set.** The current icons are minimal placeholders. A designer-grade icon set is a separate concern.
- **Offline write capability.** The offline page only serves a fallback; you can't read sessions offline. That's the correct behaviour for an agent-driven interface, but if the requirement changes, the SW has hooks for it.

---

## Reviewer FAQ

**Q: Why a 6-digit PIN and not longer?**
A: A 6-digit PIN is short enough to type on a phone in under 5 seconds. Brute-force is mitigated by the network layer (Tailscale / LAN, not public internet). The PIN is per-installation and can be regenerated any time. For a developer tool on a trusted tailnet, this is the right tradeoff.

**Q: Why PWA and not Electron/Tauri?**
A: PWA installs from the browser with one tap — no app store review cycle, no platform-specific binaries. The runtime is still the same `pi-web` Node.js binary; the PWA is an additional install surface. A native shell would re-introduce the desktop-client problem this whole project tries to avoid (one codebase per platform, manual distribution).

**Q: Why a runtime password file and not `globalThis`?**
A: Next.js 16 runs the proxy middleware in a separate worker process from the route handlers. `globalThis` is not shared across processes. The file is the smallest cross-process bridge that works.

**Q: Why is the dev server forced to reload after registering a SW?**
A: If a previous `npm start` registered a SW, that SW intercepts `/_next/static/*` and serves production chunks to the dev page, causing "module factory is not available" errors. The `PwaRegistration` dev-mode branch unregisters the stale SW and reloads.

**Q: Does this break `PI_WEB_PASSWORD`?**
A: No. `PI_WEB_PASSWORD` is the highest-priority password source. If it's set, the auto-PIN is never generated and Basic Auth still works as before.

**Q: Why is the QR token only 30s?**
A: Long enough to scan, short enough that a photographed QR is useless within seconds. Single-use prevents replay.

**Q: Why `app/manifest.ts` instead of a static `manifest.webmanifest`?**
A: Next.js metadata route — generates the manifest at build time with proper headers (`Cache-Control: public, max-age=0, must-revalidate`) and TypeScript safety on the schema. The static file approach doesn't get type checking on the manifest shape.

---

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


huangyxi:
The PWA feature had already been implemented in commit 68853091f5bf92b113f0676326526d31a317f5e1

agegr:
Thanks for the work, but this PR is too large and mixes unrelated concerns. Tailscale discovery belongs to the deployment layer, not Pi Web, and the pairing implementation currently has security and correctness issues.

I suggest closing this PR and, if needed, submitting a much smaller PR for QR pairing based on an explicitly configured public URL.

PR #561, opened 2026-08-20
fix: prune removed models from enabled scope

### LINKED ISSUE #560: Model dialog doesn't clean up `enabledModels` in settings.json when a connector/models are removed → persistent 'No models match pattern' warnings
## Summary

Removing a connector (or its models) in the pi-web model dialog does **not** clean up the stale entries in the global `enabledModels` setting (`~/.pi/agent/settings.json`). After the connector is gone, pi-web keeps showing `Warning: No models match pattern "..."` (rendered as a **Model scope warning** banner) for every subsequent session, and the only workaround is manually editing `settings.json`.

## Environment

- pi-web: **0.8.9** (`@agegr/pi-web`)
- pi-coding-agent: **0.84.2**
- macOS

## Repro steps

1. Have a connector configured whose models are part of the enabled scope. This persists entries like these into `~/.pi/agent/settings.json`:

   ```json
   "enabledModels": [
     "vercel-ai-gateway/poolside/laguna-s-2.1-free",
     "vercel-ai-gateway/alibaba/qwen3.7-flash"
   ]
   ```

2. In pi-web's model dialog, **remove the connector** (it disappears from `models.json` and the catalog).

3. Reload / open a session. The UI now shows:

   ```
   Model scope warning
   No models match pattern "vercel-ai-gateway/poolside/laguna-s-2.1-free"
   No models match pattern "vercel-ai-gateway/alibaba/qwen3.7-flash"
   ```

4. `~/.pi/agent/settings.json` **still contains the two stale `enabledModels` entries** — the dialog never prunes them.

## Root cause

- The model dialog's write path (`PUT /api/models-config`) only writes `~/.pi/agent/models.json` (provider/model config). It sanitizes empty model IDs and normalizes `cost`, but it **never touches `settings.json`**.
- pi-web **never calls `setEnabledModels`** — grep of the entire build (server + frontend) shows zero occurrences. The only writer of `enabledModels` is the pi **CLI** `/models` selector (`interactive-mode.js` → `settingsManager.setEnabledModels(...)`), and even that selector deliberately **re-injects** no-match patterns back into the enabled list (`ids.push(diagnostic.pattern)`), so it doesn't prune them either.
- `GET /api/models` reads `settingsManager.getEnabledModels()` and feeds the patterns into `resolveModelScopeWithDiagnostics()`. Every pattern with no catalog match becomes a `no-match` diagnostic, which is returned as `modelScopeWarnings` and rendered as the warning banner.

So: removing a connector in pi-web removes the models from `models.json`, but the stale `enabledModels` patterns in `settings.json` are never pruned, and scope resolution keeps reporting them on every load.

## Expected behavior

Removing a connector/models in the model dialog should also prune the corresponding `enabledModels` entries from `settings.json` (e.g. by calling `settingsManager.setEnabledModels(...)` with the filtered list), so no stale-pattern warnings appear afterwards.

## Suggested fix (any of these)

1. **Prune on connector removal:** when `PUT /api/models-config` saves the provider list, intersect the existing `enabledModels` with the new model set and persist the pruned list.
2. **Prune in scope resolution:** in `GET /api/models`, drop `no-match` patterns instead of surfacing them as warnings — or filter them out of the enabled set before persisting.
3. **CLI parity:** make the `/models` selector stop re-adding `no-match` patterns, or surface an explicit "remove stale entries" action.

## Notes

- Workaround for users today: remove stale entries from `enabledModels` in `~/.pi/agent/settings.json`.
- The same stale-entry problem can be triggered from the CLI (connector removed elsewhere), so a fix in scope resolution (#2) would cover both entry points.

y4ho0:
## Summary

- prune stale canonical `provider/modelId` entries from global `enabledModels` when the model dialog removes those configured models
- support exact entries with thinking-level suffixes while preserving globs, fuzzy or bare model ids, and unmatched future references
- clear `enabledModels` when every exact entry was removed
- invalidate the model-list cache after the settings update so stale warnings cannot be re-cached

## Why

Saving the model dialog updated `models.json` but never updated `settings.json`. Exact enabled-model entries for a deleted connector therefore produced a persistent “No models match pattern” warning in every later session.

The cleanup compares the normalized configuration before and after the save. It only removes a canonical entry when that exact model existed in the old config and is absent from the new one. User-authored patterns that may target built-in or future models are left untouched.

## Validation

- focused model-config and pruning tests: 8 passed
- full suite excluding the environment-dependent proxy test: 590 passed
- changed-file ESLint and `git diff --check`: passed
- the full suite’s sole failure is an HTTP proxy tunnel `502`, reproduced unchanged on `origin/main`
- TypeScript checking is currently blocked by the existing `ThemeColor` assertion in `lib/rpc-manager.ts`, also reproduced on `origin/main`

Closes #560

### DIFF
--- a/app/api/models-config/route.ts
+++ b/app/api/models-config/route.ts
@@ -1,4 +1,7 @@
 import { NextResponse } from "next/server";
+import { getAgentDir, SettingsManager } from "@earendil-works/pi-coding-agent";
+import { pruneRemovedEnabledModels } from "@/lib/enabled-model-pruning";
+import { invalidateModelsCache } from "@/lib/models-cache";
 import { readModelsConfig, writeModelsConfig } from "@/lib/models-config-store";
 
 export const dynamic = "force-dynamic";
@@ -10,8 +13,16 @@ export async function GET() {
 export async function PUT(req: Request) {
   try {
     const body = await req.json() as Record<string, unknown>;
-    writeModelsConfig(body);
-    return NextResponse.json({ success: true });
+    const previousConfig = readModelsConfig();
+    const nextConfig = writeModelsConfig(body);
+    const settings = SettingsManager.create(process.cwd(), getAgentDir());
+    const prunedEnabledModels = await pruneRemovedEnabledModels(
+      settings,
+      previousConfig,
+      nextConfig,
+    );
+    if (prunedEnabledModels > 0) invalidateModelsCache();
+    return NextResponse.json({ success: true, prunedEnabledModels });
   } catch (error) {
     return NextResponse.json({ error: String(error) }, { status: 500 });
   }
--- a/lib/enabled-model-pruning.test.mjs
+++ b/lib/enabled-model-pruning.test.mjs
@@ -1,0 +1,104 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+import { createJiti } from "jiti";
+
+const jiti = createJiti(import.meta.url);
+const {
+  pruneRemovedEnabledModelPatterns,
+  pruneRemovedEnabledModels,
+} = await jiti.import("./enabled-model-pruning.ts");
+
+const previousConfig = {
+  providers: {
+    acme: {
+      models: [
+        { id: "removed" },
+        { id: "nested/model" },
+        { id: "openrouter/model:exacto" },
+        { id: "remaining" },
+      ],
+    },
+  },
+};
+
+const nextConfig = {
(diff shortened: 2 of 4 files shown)


agegr:
`enabledModels` is a pi global setting and is not managed by pi-web. I don't think saving `models.json` should implicitly modify it; stale entries should remain a warning or be cleaned up explicitly by pi itself.

PR #571, opened 2026-08-21
chore: sync Bun lockfile with current dependencies

To3akaRin:
## Summary

- Regenerate `bun.lock` from the existing lockfile with Bun 1.4.0.
- Align Bun resolutions with the dependency versions and ranges already present in `package.json`.

## Why

The lockfile on `main` still records older direct dependency metadata. As a result, a frozen Bun lockfile check exits with `lockfile had changes, but lockfile is frozen`.

## Validation

- `bun install --frozen-lockfile`
- `npm test` (587 tests passed)
- `npm run lint`

Lockfile-only change; no source or public API changes.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,17 +5,17 @@
     "": {
       "name": "pi-web",
       "dependencies": {
-        "@earendil-works/pi-agent-core": "0.84.1",
-        "@earendil-works/pi-ai": "0.84.1",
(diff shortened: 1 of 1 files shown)


agegr:
Thanks for the update. We have decided to use package-lock.json as the only lockfile and removed bun.lock from main, so this PR is no longer needed.

Issue #586, opened 2026-08-23
CI + test harness: 35 pre-existing failures on clean main, no automated test gate


Nuctori:
## Summary

The repo ships a `npm test` script (`node --experimental-strip-types --test "app/**/*.test.mjs" …`) but has **no CI** and **35 pre-existing test failures** on a clean `main`. The failures are not product bugs — they are test-harness/environment issues that never get caught because nothing runs the suite automatically. This makes "does my change break tests?" unanswerable, which is a governance gap.

This issue (1) documents the failures with evidence, (2) asks for CI to run the suite on every push/PR, and (3) asks for first-class E2E coverage of the session-history paths (#509 / #555).

## Evidence

On a clean upstream checkout (`2a6e537` Release v0.8.9, no local changes):

```
npm test
ℹ tests 587
ℹ pass  549
ℹ fail   35
```

The 35 failures come from two root causes, neither of which touches application logic:

### 1. Component/hook tests fail with `useI18n must be used inside I18nProvider`
Affected: `components/MermaidBlock.test.mjs`, `components/CodeBlock*`, `components/MessageView.test.mjs`, `components/AppShell.file-viewer-state.test.mjs`, `components/ChatInput.test.mjs`, `components/ExtensionStatusBar.test.mjs`, `components/ExtensionWidgets.test.mjs`, `components/TurnWrittenFiles.test.mjs`, `hooks/useAgentSession.test.mjs`, …

Tests render with `renderToStaticMarkup(<I18nProvider><Component/></I18nProvider>)` and the provider **is** wrapped (e.g. `MermaidBlock.test.mjs:21-25`). The error is thrown because `useContext(I18nContext)` returns `null` — the provider and the consumer resolve **two different `I18nContext` instances**.

Root cause: under `node --experimental-strip-types`, `createJiti({ moduleCache: false })` re-instantiates modules, so `react` / `I18nContext` loaded by the test entry and by the imported `.tsx` are separate module instances. The context created by the test's `I18nProvider` is not the one `useI18n()` reads. This is a harness incompatibility, not a code defect.

### 2. `lib/project-command-env.test.mjs` fails on Windows PATH separator
`direct bash updates the platform PATH key` expects `PATH: '...\bin;/usr/bin'` (Windows `;`) but gets `:` (POSIX). Environment-specific assertion, not a logic bug.

Both classes fail identically on the unmodified baseline, so they are pre-existing and unrelated to any feature PR.

### A. Add CI (GitHub Actions)
[...]

PR #587, opened 2026-08-23
fix: bound session history load to a tail window (fixes #509, #555)

### LINKED ISSUE #509: Feature request: conversation management & chat archive (对话管理和对话存档)
### Feature request: conversation management & chat archive

**Motivation**

As pi-web is used over time, the sidebar session list grows large and it becomes hard to find, organize, or keep a backup of past conversations. This request covers two things: (1) better conversation management in the UI, and (2) a way to archive / export chat history.

**1. Conversation management**

- Search & filter sessions by keyword, project, or time range
- Batch operations: select multiple sessions → delete, export, or move into a folder
- Sort options (last active / project / model)
- Complements the existing folder/category work in #458

**2. Chat archive / export**

- Export a single session (or all sessions) to Markdown / HTML / JSON — mirroring pi CLI's `/export` and `/share` capabilities
- Optional scheduled auto-archive: keep the live list trimmed to the last N days, move older sessions to an archive
- Preserve tool calls, thinking blocks, and metadata so archives can be re-imported or audited later
- A "download backup" button in the session detail view would help users running pi-web on a remote / headless box

**Use case**

A typical day involves many short conversations; without search and batch export, useful context gets buried and can be lost when the machine is cleaned up or the session store is reset.

Thanks for the great project!

### LINKED ISSUE #555: Performance: page session detail history instead of returning the full active branch
## Problem

Related to #500.

Opening or refreshing a long saved session calls `GET /api/sessions/:id`. The route reads the entries, builds the entire active-branch context, and returns it in one JSON response. The chat UI initially renders only the most recent `VISIBLE_PAGE_SIZE` messages, but the browser still downloads and JSON-parses the full active history before the session becomes usable.

`deferThinking` and `deferMedia` reduce some payload categories, but ordinary text, tool-result text, and historical metadata still scale with the complete active branch. Large sessions therefore make initial open and refresh disproportionately slow, especially on mobile or constrained links.

## Expected behavior

Initial session open should transfer and parse only the recent history needed for the first viewport. Older history should load on demand without changing session-tree, fork, edit, or JSONL semantics.

## Proposed direction

- Add a backward-compatible cursor/page mode to the session-detail API that returns the newest logical history page plus `hasMore` and a stable cursor.
- Use active-branch entry IDs rather than unstable array offsets, and avoid splitting a coherent assistant/tool-result turn when practical.
- Have the existing top-of-history lazy-load interaction request older pages and preserve scroll position.
- Keep the current full response behavior for callers that do not opt into pagination until the frontend migration is complete.
- Consider deferred retrieval for exceptionally large collapsed tool-result text as a separate follow-up, rather than silently truncating history.

## Acceptance criteria

- A long session opens with only its tail page transferred and parsed initially.
- Loading older pages neither duplicates nor skips active-branch messages.
- Branch navigation, forking, editing, compaction display, and session export retain their current behavior.
- Existing API consumers remain compatible during the migration.

Nuctori:
## Summary

Bounds how much of a session is loaded when a session detail is opened, so deep/long sessions no longer crash or stall.

- **#500 (stack overflow / HTTP 500 on open)**: `BranchNavigator` walked the active path with recursion over an arbitrarily deep chain; rewritten as explicit-stack iteration. `lib/session-reader` already slices the chain, so the SDK recursion that previously overflowed is no longer on the open path.
- **#555 (full-history transfer / slow open)**: detail + context APIs returned the entire entry forest. They now return only a `tail` window of the active branch, and support upward pagination via `?before`, so opening a 5000-message linear session transfers ~50 messages instead of 5000.
- **extra crash found via real-browser E2E**: long sessions with a string `assistant` `content` (not a block array) hit `TypeError: message.content.map is not a function` in `entryToUiMessage`. Guarded with an `Array.isArray` check.
- **Codex review fixes** (P1/P2 on the earlier branch state): restore multi-root branch detection in `hasBranch` (iterative rewrite had dropped the `nodes.length > 1` check); expand the rendered window so prepended (older) pages stay visible and the "load earlier" sentinel shows when the initial tail is a truncation.

## Reproduce

A session with >3000 messages and no branches (a single linear chain). Open it:
- before: `GET /api/sessions/:id` → `HTTP 500`. List page is fine.
- after: detail returns `200` with `context.messages.length === 50`; scrolling up fetches `?before=<oldest>` pages without duplicating the boundary.

## Changed

- `lib/session-reader.ts`: `sliceActiveBranch` (+`excludeLeaf` for page dedupe), `buildSessionContext` gains `tail`/`excludeLeaf`; `entryToUiMessage` guards string `content`.
- `app/api/sessions/[id]/route.ts`, `app/api/sessions/[id]/context/route.ts`: `?tail` (default 50, capped 1000, NaN-safe) and `?before` pagination params.
- `hooks/useAgentSession.ts`: `loadContext(sid, leafId, before?)` prepends older pages.
- `components/ChatWindow.tsx`: top sentinel triggers `loadContext` for earlier history; rendered window expands to keep prepended pages visible.
- `components/BranchNavigator.tsx`: `buildActivePath`/`hasBranch` recursion → explicit stack (with multi-root check).
- Tests: `lib/session-reader.pagination.test.mjs`, `app/api/sessions/detail-route.test.mjs`, `app/api/sessions/context-route.test.mjs`, `components/BranchNavigator.test.mjs` (6000-deep chain + multi-root).
- `package.json`: version 0.8.9 → 0.8.10.

## Verification

- `npm test`: with the harness fix from #588 applied, 604 tests / 592 pass / 9 fail. The 9 remaining failures are platform-conditional (Windows symlink / PATH separator) or pre-existing brittle source-structure assertions — no regression from this change.
- `tsc --noEmit` clean; ESLint clean on changed files.
- Real-browser E2E (5000-message linear session): detail `200` + 50 messages, `?before` pages without duplicate, no console/page errors (run manually; the CI E2E job lives in the CI PR).

## Note

- `package-lock.json` and `AGENTS.md` are intentionally not part of this PR.
- **Depends on #588** (test-harness fix): without it, the component tests throw `useI18n must be used inside I18nProvider` and the suite shows 35 failures on a clean checkout. Merge #588 first.

Closes #509. Closes #555.


Nuctori:
Update: this branch now also fixes the component-test harness (see #586). React, react-dom/server, and the i18n provider are loaded through the same jiti instance the components use, and useI18n is imported via the @/ alias so I18nContext resolves to a single module — previously two React instances made every component test throw 'useI18n must be used inside I18nProvider'.

Suite: 603 tests, 591 pass, 9 fail (was 565/35). Remaining 9 = platform-conditional (symlink, Windows PATH separator) + pre-existing brittle source-structure assertions, not regressions. No product code changed by the harness fix.

Nuctori:
Split out the test-harness fix into its own PR (#588) to keep this one scoped to #509/#555. Reverted the harness commit from this branch.

Nuctori:
Addressed all three P2 items in 7c9b27f:

1. Reveal prepended history pages — fixed. ChatWindow now keeps the rendered window at least as large as the loaded message count (Math.max(visibleCount, messages.length)), so prepended older pages are no longer sliced off the top. Also shows the 'load earlier' sentinel when the window is full (the initial tail is a truncation), so the first page can always be fetched.
2. Preserve multiple root branches — fixed. hasBranch iterative rewrite had dropped the 
odes.length > 1 check; restored it. Sessions branched from the first message (multiple roots) are now correctly reported as branched. Added a test for this case.
3. Lockfile for Playwright — removed. playwright was only added locally for E2E probing and does not belong to this fix; dropped it from package.json (it was never in the upstream lockfile, so there is nothing to regenerate).

Nuctori:
PR cleaned up: restored the mermaid devDependency (a previous edit had accidentally replaced it with a duplicate postcss entry) and dropped the temporary local e2e probe scripts (2e_browser.mjs etc.) from the tree. Body now states the dependency on #588 for the component tests to pass on a clean checkout.

PR #588, opened 2026-08-23
test: fix component-test harness dual React instance (35 -> 9 failures)

Nuctori:
## Summary

Fixes the component-test harness so React component tests actually run. This is the harness half of #586 (the other half — adding CI — is still wanted).

Every component test that renders with `renderToStaticMarkup(<I18nProvider>…)` was failing with `useI18n must be used inside I18nProvider`, even though the provider was wrapped correctly. Root cause: the test entry imported `React` natively (`import React from "react"`) while jiti loaded the `.tsx` tree, so two React instances and two `I18nContext` modules existed; `useContext` returned `null`. The `useI18n` import also used a `.tsx` suffix that jiti resolved to a different module cache entry than the components' `@/hooks/useI18n`, producing a second distinct `I18nContext`.

## Changed

- `components/{MermaidBlock,ChatInput,ExtensionStatusBar,ExtensionWidgets,MessageView,TurnWrittenFiles}.test.mjs`: load `react`, `react-dom/server`, and `I18nProvider` through the same jiti instance the components use, and import `useI18n` via the `@/` alias (matching component imports) so `I18nContext` resolves to one module instance. No product code touched.

## Verification

- `npm test`: 603 tests, **591 pass, 9 fail** (was 565/35 before this fix on the same branch, 565/35 on a clean `main` baseline).
- Remaining 9 failures are **not harness defects**: `lib/directory-browser` (symlink `EPERM` on Windows, passes on Linux), `lib/project-command-env` (Windows `PATH` separator `;` vs `:`), `components/AppShell.file-viewer-state` (brittle source-comment-block assertions), `hooks/useAgentSession` (4 brittle internal-variable regex assertions). These pass on Linux CI or are pre-existing test debt, out of scope here.

## Note

This PR is scoped to the test harness only. The pagination feature work (the session-history tail/before changes that Codex reviewed on the earlier branch state) lives in PR #587; the Playwright E2E example for CI lives in the CI PR. No product code in this PR.

Relates to #586.


Nuctori:
Branch rebuilt from main so this PR is scoped to the harness fix only (the earlier branch state accidentally carried the pagination feature from #587 as well). Codex's P1/P2 notes about the detail route (pp/api/sessions/[id]/route.ts), lib/session-reader.ts, ChatWindow.tsx, and the playwright lockfile belong to PR #587, not this harness PR — they are addressed there (see #587). This harness PR touches no product code and no package.json.

PR #593, opened 2026-08-25
fix(notices): render multi-line notices (e.g. /acp panel) without truncating

ranxianglei:
## Problem

Extension `ctx.ui.notify(...)` messages that span multiple lines are truncated to a single line in the notice shelf. The clearest example is the **billion-context-pi `/acp` status panel** (a ~15-line monospace block: context usage, sent-to-LLM, nudge state, blocks), which shows up as a one-line fragment ending in "…".

## Root cause

`NoticeShelf` (`components/ChatWindow.tsx`) rendered **every** notice in a fixed `height: 60` box with `whiteSpace: "nowrap"` + `textOverflow: "ellipsis"`. Any message containing newlines was collapsed onto one line and clipped. Notices also auto-dismiss after `NOTICE_VISIBLE_MS` (5s) — too short to read a multi-line panel even if it weren't clipped.

The server side (`lib/rpc-manager.ts` `notify`) and the event wiring (`hooks/useAgentSession.ts` `case "notify"` → `addNotice`) were already correct — the panel reaches the browser; only the rendering clipped it.

## Fix

- **`components/ChatWindow.tsx`** — detect multi-line notices (`message.includes("\n")`) and render them with `whiteSpace: "pre-wrap"`, a monospace font, auto height (scrollable, capped at 420px), and top alignment. Single-line notices keep the existing toast styling unchanged.
- **`hooks/useAgentSession.ts`** — add an optional `visibleMs` to `NoticeItem`; multi-line notices stay visible for 30s (`MULTI_LINE_NOTICE_VISIBLE_MS`) instead of 5s so the panel can be read. The dismiss timer now honors the oldest notice's `visibleMs`.
- **`components/ChatWindow.notices.test.mjs`** — add a regression test for the multi-line rendering path.

## Verification

- `tsc --noEmit` — clean
- `npm run lint` — clean
- `npm test` — 588 pass, 0 fail

### DIFF
--- a/components/ChatWindow.notices.test.mjs
+++ b/components/ChatWindow.notices.test.mjs
@@ -13,3 +13,8 @@ test("renders temporary notices once at the top center of the chat column", () =
     /position: "absolute",\s*top: 12,\s*left: 0,\s*right: isMobile \? 0 : CHAT_MINIMAP_WIDTH,[\s\S]*?justifyContent: "center",[\s\S]*?<NoticeShelf notices=\{notices\} floating \/>/,
   );
 });
+
+test("renders multi-line notices (e.g. /acp status panel) with pre-wrap monospace instead of truncating to one line", () => {
+  assert.match(source, /const isMultiLine = notice\.message\.includes\("\\n"\);/);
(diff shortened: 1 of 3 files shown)


agegr:
#558  supersede

PR #607, opened 2026-08-25
feat: include renamed session name in browser tab title

almendio-dev:
Extends the browser tab title to include the user-defined session name,
when one exists, alongside the workspace directory name.

Title patterns:
- {sessionName} - {cwd} - Pi Web
- {sessionName} - Pi Web
- {cwd} - Pi Web
- Pi Web (fallback)

The session name is read from the latest session catalog entry so the
title updates immediately when the user renames the selected session
in the sidebar.

Verification:
- node_modules/.bin/eslint components/AppShell.tsx

Note: npm run lint still reports 15 pre-existing React Compiler
memoization errors in other files; they are unrelated to this change.


dingxiaobo:
修改前的 title 非常适合我，

如果要合入此 PR，希望在设置中，把 title 做成可配，能让我继续用老款

almendio-dev:
> 修改前的 title 非常适合我，
> 
> 如果要合入此 PR，希望在设置中，把 title 做成可配，能让我继续用老款

@dingxiaobo  I updated the PR to have opt-out settings.

agegr:
Thanks for the contribution. I understand the intended use case: opening multiple browser tabs for different sessions under the same cwd, where the workspace-only title makes those tabs difficult to distinguish.
However, I don’t think we should add this feature at this time. Pi Web already supports switching between sessions within the same workspace, so this mainly benefits the narrower workflow of keeping multiple tabs open for the same cwd. It also only helps sessions with an explicitly stored name, while adding a new preference, settings UI, persisted state, and title synchronization logic.
Given the limited scope of the use case, I don’t think the additional product and maintenance complexity is justified.

almendio-dev:
Thanks for elaborating @agegr . One more use case is automated time tracking that's aware of the window / tab title. Probably another edge case, but if you see the need in the future, feel free to reopen the PR. 
Thanks for the great app!

PR #618, opened 2026-08-26
feat: full-text search over past conversations

FelixWang47831:
# PR description

The sidebar can browse sessions per project, but finding a conversation by what
was said in it means grepping `~/.pi/agent/sessions` by hand. A magnifier next to
**New** / **Refresh** (or `Ctrl`/`Cmd`+`Shift`+`F`) replaces the session list
with a debounced full-text search over the stored session JSONL files: results
are grouped per session with highlighted snippets, scopable to one project and
optionally including tool output, and clicking one opens that session at that
message, scrolls to the first keyword occurrence and highlights every occurrence
in it. Adds `GET /api/sessions/search` and `?from=<entryId>` on the context
route; i18n for `en`/`zh-CN`/`zh-TW`; no new dependencies.

image

- **No index** — `lib/session-search.ts` streams the files on every keystroke,
  takes its candidates from `listAllSessions()` (so project scoping matches the
  sidebar and no request can escape the sessions directory), and is bounded by a
  file cap, an 8s budget and a per-session hit cap that it reports back so the UI
  can say "narrow the search".
- **Jumps stay contiguous with the leaf** — the chat only pages upward, so
  `?from=` returns everything from the target entry through the leaf, and refuses
  the jump (unknown entry, other branch, too far back) instead of shipping an
  isolated window.
- **The highlight uses ranges, not markup** — `<mark>` would split text nodes
  React owns, so `lib/text-highlight.ts` uses the CSS Custom Highlight API
  (Chrome/Edge 105+, Safari 17.2+, Firefox 140+; elsewhere it degrades to the row
  flash), and `lib/text-match.ts` is new so client and server share one matcher.

45 tests added, `tsc --noEmit` and `npm run lint` clean, and the remaining
`npm test` failures also fail on an unmodified checkout (Windows CRLF and `PATH`
assertions).


agegr:
Thanks for the contribution. I've implemented a simpler version focused on conversation text search and basic message navigation, so I'm closing this PR in favor of that implementation to keep the initial scope small.

PR #621, opened 2026-08-27
feat: integrate Cursor subscription models

soffchen:
## Summary

- add Cursor OAuth polling and account-scoped model discovery
- stream Cursor Connect-RPC responses directly through the pi `streamSimple` provider extension, including tool continuation, cancellation, proxy support, and session cleanup
- register Cursor through provider-neutral runtime helpers and expose it in the existing auth and model UI

## Verification

- `npm run lint`
- `npm test` (882/882 tests passed)
- real Cursor OAuth and model discovery smoke test (112 usable models)
- real model request returned `OK`


soffchen:
image
Tested locally with a real Cursor Ultra account. OAuth login completed successfully, account-scoped model discovery returned 112 usable models, and an end-to-end model request completed with the expected OK response. The provider is available in the model selector and works as expected. Type checking, linting, and the full test suite also pass (882/882).

agegr:
This integration should be implemented and distributed as a Pi plugin. pi-web is a web UI for Pi, and bundling Cursor-specific OAuth, model discovery, and Connect-RPC support here would make this project responsible for maintaining compatibility with Cursor's protocol.

Please move the Cursor provider implementation into a plugin using Pi's existing provider extension APIs. Any generic changes needed in pi-web to support plugin-provided providers can be proposed separately.

PR #617, opened 2026-08-26
test(e2e): add GitHub Actions CI + Playwright E2E suite for user journeys

Nuctori:
Closes #599 — example implementation of the CI + E2E proposal (items A and C of the issue; item B / harness fix already landed via #588).

## What this adds

**1. GitHub Actions CI (`.github/workflows/ci.yml`)**

- Four jobs on push(main) + pull_request: lint / test / build / e2e
- Node 22 (aligned with `engines >=22.19`), `cache: npm`
- e2e job boots the dev server, waits for readiness, runs the full E2E suite

**2. Script-style Playwright E2E suite (`e2e/`, zero config)**

| Suite | Coverage |
| --- | --- |
| `run.mjs` | 5000-message linear session: tail window + sentinel + no crash (#509/#555 baseline) |
| `j0-shell-nav` | App shell renders, theme control present |
| `j1-session-lifecycle` | Long-session tail enforcement (API-level assertion: context ≤ tail cap, oldest message excluded), branch session renders active path |
| `j2-dialogue` | Markdown / code block / tool-call rendering |
| `j3-model-agent` | Model/tool UI presence |
| `j4-files-project` | Files/project UI smoke |
| `j5-settings-ops` | Settings/plugins/skills UI smoke |
| `j6-api-health` | **Hard 500-gate**: detail/context/list return 200 + correct JSON shape; invalid id, bad leafId, and path traversal must NOT 500 |

Helpers isolate all fixtures under `~/.pi/agent/sessions/e2e-*` with cleanup; no `playwright.config.ts`, each spec is a plain runnable script.

**Coverage honesty**: J1/J6 (+ run.mjs) are hard assertions; J2/J3 are smoke-level; J4/J5 are soft-presence checks. Sending/streaming/fork interactions need a real agent backend and are out of scope for this first pass — listed as follow-up work on #599.

## Why

# 509 (deep-chain stack overflow) and #555 (full-history transfer) were exactly the class of regression unit tests miss. This gate now blocks them automatically on every PR — including bulk/AI-assisted contributions where "looks right, crashes at runtime" regressions are most likely.

## Validation

- Branch CI fully green: [run 32946773853](https://github.com/Nuctori/pi-web/actions/runs/32946773853) — lint ✅ test ✅ build ✅ e2e ✅ (the e2e job itself caught and fixed 5 real spec bugs during development, incl. a sidebar-preview false positive and a 30s session-list cache race)
- `npm test`: same pass/fail counts as upstream `main` (no product code touched)

[...]

PR #624, opened 2026-08-27
ci: add independent perf-bench job (cold/warm list, context tail=50)

Nuctori:
## What

Adds a `perf` job to CI that measures four numbers on a pristine dev server
with a synthetic 200-session + 1×5000-msg fixture, printed as JSON, with
generous order-of-magnitude gates for trend/regression detection.

- `coldListMs`       first `GET /api/sessions` (~200 fixture sessions)
- `warmListMs`       second `GET /api/sessions` (cache-hit path)
- `contextTail50Ms`  `GET /api/sessions/<big>/context?tail=50` (5000-msg)
- `pageFirstByteBytes` home HTML size (browser render deferred to e2e harness)

The gates are deliberately loose — shared CI runners are noisy — but they
catch order-of-magnitude regressions across PRs, which is what we need
before any of the upcoming L1–L4 perf changes land.

## Why

Upcoming perf PRs (fingerprint memoization, `O(delta)` tail-read, windowed
list, reverse-stream context) all need a regression gate. The numbers they
are claiming to improve (cold 70s → ~100ms, context 4.8s → 564ms on 53MB
sessions) only have meaning if there is a baseline measurement in CI to
diff against.

## Independent of #617

No `playwright`, no `e2e/helpers.mjs` dependency, no shared fixture. The
bench writes its own JSONL sessions under a tmp `PI_CODING_AGENT_DIR` that
the dev server is told about via env. Merge order between this PR and #617
is irrelevant — either can land first.

## Source changes

- 0 changes to `lib/`, `app/`, `components/`, `hooks/`
- 0 changes to `package.json` / `package-lock.json`
- 2 new files: `e2e/perf-bench.mjs` (170 lines, zero deps beyond
  `node:fs`/`node:path`/`node:os`/`node:crypto`), `.github/workflows/ci.yml`
  (adds a `perf` job alongside the existing `lint`/`test`/`build`/E2E if any)

## Audit (self)

- base: `origin/main` (28bab3c), ahead 0 before this PR
- ports: bench uses 3100; safe with anyone running 30141 locally
- cleanup: fixture dir + dev server pid are torn down in `if: always()`
- reverts: delete 2 files + remove the `perf` job; nothing else changes
- commit hygiene: 2 commits (add + drop unused import); no merge noise

## Stacked plan (each PR base=main, merge-order-free)

- P2 `perf/l1-scanner` — fingerprint memoization (`fp=(size,mtimeMs)`); new
  file `lib/session-list-scanner.ts`; ~470 lines; replaces
  `SessionManager.listAll` with `listSessionsIncremental`.
- P3 `perf/l2-tail` — `O(delta)` tail-read for grown files; ~140 lines;
  the scanner's `scanSessionFileInfo` learns a `tailUpdateForGrownFile`
  fallback for fingerprint-mismatched files.
- P4 `perf/l3-sidebar` — windowed list rendering; no `react-window` dep;
  ~85 lines; 2000+ cards drop to ~30 mounted at a time.
- P5 `perf/l4-context` — reverse-stream `loadRecentEntries` for the
  context route; +1 test fix for the mock surface; ~135 lines.

None of P2–P5 touches the other. All four stack on top of this PR for the
measurement gate, but they merge into `main` independently once green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,77 @@
+name: ci
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version: 22
+          cache: npm
+      - run: npm ci
+      - run: npm run lint
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version: 22
+          cache: npm
+      - run: npm ci
+      - run: npm test
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: actions/setup-node@39370e3970a6d050c480ffad4ff0ed4d3fdee5af # v4.1.0
+        with:
+          node-version: 22
+          cache: npm
+      - run: npm run build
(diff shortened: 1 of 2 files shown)


agegr:
Thanks for the contribution. #722 now provides CI and an isolated session-history E2E harness, so we would prefer to add performance measurements there instead of maintaining a separate setup. Closing this PR in favor of a focused follow-up using the existing harness.

PR #644, opened 2026-08-28
feat: auto-generate session title after the first completed run

looluo:
## Auto-generate session title after the first completed run

Today session titles are manual-only (top-bar "Generate title" button). This PR adds automatic titling: when a regular session's agent run settles and the session has no name, a title is generated with the session's own model and persisted.

### Design

- **Server-side trigger.** Wired into the existing `onAgentRunComplete` hook in `AgentSessionWrapper`, so background sessions, closed tabs, and multi-tab setups all behave identically — no browser needs to be watching. Subagent wrappers suppress this callback entirely.
- **Named sessions are never touched.** The trigger condition is simply "session currently has no name": once any name exists (auto-generated, manual, or set in the pi TUI) it is never overwritten. A failed attempt retries naturally on the next completed run — no state, no queues.
- **Generation logic is reused as-is** from `lib/session-title.ts` (shadow agent, session model, 90s timeout, no tool execution).
- **Concurrency.** A globalThis-scoped in-flight guard is shared between the automatic hook and the manual `/auto-name` route; a manual click during automatic generation gets an explicit 409 instead of racing two shadow agents.
- **Browser propagation.** A synthetic `session_named` event is emitted over the existing per-session SSE stream (typically still inside the 30s idle grace window) so open tabs refresh the sidebar immediately.
- **Setting.** `~/.pi/agent/pi-web.json` → `autoSessionTitle`, default `true`; atomic writes preserve unknown fields. Missing file = enabled; malformed file fail-closes the hook; the field is read at run-completion time so toggling takes effect without reloading sessions. Toggle lives in Settings → General (`GET/PUT /api/sessions/settings`).

### Commits

1. `feat: auto-generate session title after agent run completes` — hook + shared guard
2. `feat: gate automatic session titles behind a pi-web setting` — config module
3. `feat: broadcast automatic session titles over the event stream` — `session_named` SSE event
4. `feat: add automatic session title toggle to general settings` — API route + UI

### Tests

Unit tests for the trigger condition (named/dead/disabled/generate-once/retry-after-failure), the lock's exclusivity, the settings module (defaults, fail-close, field-type fallback, unknown-field preservation, round-trips), the settings route validation, and source-inspection assertions for the hook wiring, lock sharing, and event pipeline — matching this repo's existing test conventions.


agegr:
I'd prefer to keep title generation manual for now. The first-message fallback and existing “Generate title” button already cover the basic need, while automatic naming adds background model usage and lifecycle complexity. I don't think the convenience justifies adding this to Pi Web's default behavior yet.

PR #667, opened 2026-08-31
feat(ui): add __piBatchAsk__ batch questionnaire dialog + code-fence highlighting in dialog titles

YsDirector:
## Summary

Two small, backward-compatible extension-UI enhancements:

### 1. `__piBatchAsk__` batch questionnaire dialog
When an extension sends an `input` request whose `title` is a JSON envelope
`{ "__piBatchAsk__": 1, review?: boolean, questions: [...] }`, the client now
renders a tabbed batch questionnaire instead of a plain input:
- One tab per question (`select` / `confirm` / `input` / `editor`), with
  custom free-text entry for `select` when `allowOther !== false`.
- Optional Review tab (guards against accidental submit when `review: true`).
- Replies with `{ value: JSON.stringify({ answers }) }`.
- Fully backward compatible: any non-envelope `input` renders exactly as before.

This lets a single extension call (e.g. `ask_question`) present a multi-question
form to the user in one dialog, instead of a series of sequential prompts.

### 2. Code-fence highlighting in extension dialog titles
The extension dialog title is now split: the first line stays pinned in the
header, and the remainder (e.g. a fenced shell command) renders inside the
scrollable body with ```sh / ```bash blocks highlighted. Makes long prompts
(like permission reviews showing the exact command) readable and scannable.

## Files
- `components/ChatWindow.tsx` — BatchAskDialog, envelope detection, title split + highlight
- `lib/types.ts` — `BatchAskEnvelope` / `BatchAskQuestion` / `BatchAskAnswer`
- `lib/i18n/messages/{en,zh-CN,zh-TW}.ts` — new strings

## Verification
- tsc clean, 844/844 tests pass


agegr:
Closing this PR because the batch-questionnaire implementation introduces a client-specific protocol by embedding JSON in `input.title`. `__piBatchAsk__` is not part of Pi’s upstream UI API, so extensions using it would depend on Pi Web interpreting an ordinary input title as a structured form.

Pi already provides an upstream `questionnaire` example that registers a tool, accepts multiple questions, and collects answers through `ctx.ui.custom()`. That is the direction I would prefer here: one tool call containing `questions[]`, with all answers returned together. A step-by-step implementation can use sequential `select`/`input` calls without additional model round trips; a grouped questionnaire can use a single custom UI. The upstream example currently requires TUI mode, so adapting and validating it for Pi Web would still be necessary.

The current implementation also has correctness issues, including the mismatched envelope key, answers persisting across different requests, and editor/custom-input state producing unexpected submission behavior.

The dialog-title rendering improvement is independently useful and can be proposed in a separate PR with focused tests.

PR #676, opened 2026-08-31
fix: render extension dialogs inline with keyboard nav, stacking, and expiry (#462)

### LINKED ISSUE #462: Extension UI dialogs: full-window overlay hides chat context; single-dialog limit breaks select→input flows; timeout/expiresAt unimplemented
## Summary

pi-web renders extension dialogs (`ctx.ui.select()` / `ctx.ui.confirm()` / `ctx.ui.input()` / `ctx.ui.editor()`) with a UX that diverges systematically from the pi TUI. These methods are part of the public ExtensionAPI ([extensions docs](https://pi.dev/docs/latest/extensions), delivered over the RPC sub-protocol `extension_ui_request`), so **every** third-party extension that relies on them (permission confirmations, interactive configuration, questionnaires, credential prompts, etc.) is affected on pi-web — this is a platform-level gap, not specific to any one extension.

## Environment

- `@agegr/pi-web` 0.8.7 (also on main / 0.8.8-beta.1)
- `@earendil-works/pi-coding-agent` 0.84.x, RPC mode

## Current behavior (`components/ChatWindow.tsx` → `ExtensionDialog`)

1. **Full-window overlay hides the chat context.** The dialog is rendered as `position:absolute; inset:0; zIndex:90; background:rgba(0,0,0,0.18)` — the entire chat area is dimmed and covered while a dialog is open, so the user cannot re-read the conversation while answering. Questions like "which approach?" or "what does this error say?" usually have the answer in the context now hidden behind the overlay; the user must answer from memory. Clicking the overlay does nothing (no cancel, no dismiss). The TUI renders the same dialogs as an overlay that does not block the conversation and supports Esc.

2. **Single-dialog state cannot express composed flows.** The dialog is one state slot in `useAgentSession` (`setExtensionDialog`), and the RPC sub-protocol carries one dialog at a time. But "option list + fallback free-text input" (select followed by input) is a common composition pattern for TUI extensions. On pi-web this becomes two disconnected sequential modals, and cancelling the second discards the context of the first.

3. **`timeout` / `expiresAt` are defined in the protocol but never implemented.** Both fields exist in `lib/types.ts` (`ExtensionUiRequest`), yet the client ignores them entirely. Extensions cannot express "auto-cancel if nobody answers", which diverges from TUI behavior (and the daemon-side `extensionDialogsTimeoutMs` safety valve does not cover this client-side affordance).

4. **`select` has no keyboard support.** Options are mouse-only buttons; no arrow-key navigation and no Esc handling (input/editor do handle Escape, select does not).

## Expected behavior

- Dialogs no longer hide the chat context: render them as **inline cards in the conversation transcript** (a design already validated by a sibling pi GUI project, where dialog cards render inline in the session and resolve with the user's actual answer), or at minimum lighten the overlay and allow click-outside-to-cancel.
- Support dialog sequences / composed flows (the select → input pattern).
- Implement the already-defined `timeout` / `expiresAt` fields.
- Keyboard navigation and Esc for `select`.

## Proposed solution (incremental, each step independently shippable)

- **P0 (one component):** add keyboard navigation + Esc to `select`; click-outside = cancel; reduce overlay opacity.
- **P1:** move dialogs from a full-window overlay to inline cards in the message stream (options + input rendered in the transcript; answers travel back over the existing `extension_ui_response` channel).
- **P2:** support a dialog stack (composed select → input flows) and implement `timeout`/`expiresAt`.

## Reference

- Upstream protocol: `extension_ui_request` in `lib/types.ts` — select/confirm/input/editor all carry `timeout`/`expiresAt`.
- Validated sibling implementation: inline dialog cards in the session transcript (jmfederico/pi-web, "Pi extension dialogs in PI WEB").
- Repro: have any extension call `ctx.ui.select()` (e.g. a question tool with a couple of options), then observe the overlay and the keyboard behavior.

mufflegyro:
Closes #462

## Summary

`ctx.ui.select()` / `confirm()` / `input()` / `editor()` dialogs were rendered as a full-window overlay that hid the chat context, ignored the `timeout`/`expiresAt` fields in the protocol, offered no keyboard support for `select`, and used a single state slot that couldn't express composed flows. This PR implements all three proposed steps (P0 + P1 + P2) from the issue.

## Changes

**P1 — Inline cards in the transcript**
- Dialogs now render as inline cards in the conversation transcript (message stream, plus the new-session state below the composer) instead of a dimming `position:absolute; inset:0` overlay. The conversation stays readable while answering; no overlay, no click-outside needed.
- Dropped `aria-modal="true"`: the card is intentionally non-modal (`role="dialog"` + `aria-label` instead).

**P0 — Keyboard navigation + Esc**
- `select` supports Arrow Up/Down (and Left/Right), Home/End, Enter to pick the highlighted option, and Esc to cancel. Hover syncs the active option.
- The global stop-agent Esc shortcut now yields when focus is inside an extension dialog (`hooks/useKeyboardShortcuts.ts`), so Esc cancels the dialog instead of aborting the run.

**P2 — Dialog stack + timeout/expiresAt**
- Unanswered dialogs queue in a stack instead of clobbering each other, so composed flows (select → input) and pipelined requests from concurrent extensions resolve in order. Entries are deduped by request id because the events stream replays pending requests on reconnect (`hooks/useAgentSession.ts`).
- Dialogs carrying `timeout`/`expiresAt` show a live countdown ("expires in Ns") and auto-cancel at the deadline, resolving the extension's await with its default value — matching the daemon-side timeout and the TUI's auto-dismiss. Stale dialogs are dropped when switching sessions.

## Verification

- `tsc --noEmit`, `eslint .`, and the full test suite (844 tests) pass.
- End-to-end with a real extension command on the running dev instance (playwright, headless chromium):
  - Keyboard select → input composed flow round-trips the user's actual answers back to the extension;
  - Esc cancels the select (notify received);
  - a 6s `timeout` shows the countdown, auto-dismisses, and notifies the extension;
  - two queued dialogs show one active card + a "N more pending" indicator and resolve sequentially;
  - a page reload (SSE reconnect) replays only one dialog card (dedupe).

## Notes

- `ExtensionCustomPanel` (custom UI method) is intentionally left as an overlay — out of scope for this issue.
- The select options list caps at `maxHeight: 260` with internal scroll (replaces the old `flex: 1` overlay cap); the card is `maxWidth: 560`, `width: 100%`.

### DIFF
--- a/components/ChatWindow.tsx
+++ b/components/ChatWindow.tsx
@@ -285,7 +285,7 @@ export function ChatWindow({ session, sessionRunning, newSessionCwd, newSessionD
     retryInfo, contextUsage, forkingEntryId,
     isCompacting, compactError, compactResult, displayModel: displayModelValue, modelSwitching, sessionStats,
     slashCommands, slashCommandsLoading, queuedMessages,
-    notices, extensionDialog, extensionCustomUi, extensionStatuses, extensionWidgets, respondToExtensionUi, sendExtensionCustomInput, setNoticePaused,
+    notices, extensionDialogs, extensionCustomUi, extensionStatuses, extensionWidgets, respondToExtensionUi, sendExtensionCustomInput, setNoticePaused,
     isAutoModelSelection,
     agentPhase,
     isNew,
@@ -302,6 +302,10 @@ export function ChatWindow({ session, sessionRunning, newSessionCwd, newSessionD
     modelsRefreshKey, chatInputRef, onBranchDataChange, onSystemPromptChange, onSystemToolsChange, onSystemInfoLoaderChange, onSessionStatsPanelOpen,
   });
   const sessionBusy = agentRunning || bashRunning;
+
+  // The active dialog is the most recent unanswered request; older ones wait
+  // in the stack and surface as each earlier one is answered or expires.
+  const extensionDialog = extensionDialogs.length > 0 ? extensionDialogs[extensionDialogs.length - 1] : null;
 
   useEffect(() => {
     if (
@@ -657,13 +661,6 @@ export function ChatWindow({ session, sessionRunning, newSessionCwd, newSessionD
         </div>
       )}
 
-      {extensionDialog && (
-        <ExtensionDialog
-          request={extensionDialog}
-          onRespond={respondToExtensionUi}
-        />
-      )}
-
       {extensionCustomUi && (
         <ExtensionCustomPanel
           request={extensionCustomUi}
@@ -718,6 +715,7 @@ export function ChatWindow({ session, sessionRunning, newSessionCwd, newSessionD
               </div>
             </div>
             {chatInputElement}
+            <ExtensionDialogCardList dialogs={extensionDialogs} onRespond={respondToExtensionUi} />
             <ExtensionStatusBar statuses={extensionStatuses} widgets={extensionWidgets} />
           </div>
         </div>
@@ -944,6 +942,8 @@ export function ChatWindow({ session, sessionRunning, newSessionCwd, newSessionD
               />
             )}
 
+            <ExtensionDialogCardList dialogs={extensionDialogs} onRespond={respondToExtensionUi} />
+
             <div ref={promptAnchorSpacerRef} aria-hidden="true" />
 
             <div ref={messagesEndRef} />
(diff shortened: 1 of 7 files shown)


agegr:
Closing this PR in favor of the existing #612 changes and https://github.com/agegr/pi-web/commit/2356904df7a504efa0818fdcecf3f18a31eeefa1 on main.

#612 already provides collapsible dialogs that keep the composer and Stop accessible, plus server-driven dialog closure on timeout/cancellation. The new commit adds arrow-key and Home/End navigation with visible native button focus, scrolls the focused option into view, handles Escape inside the dialog without aborting the agent, and displays a localized countdown from the server's expiresAt value. Expiry remains owned by the server through extension_ui_closed.

The inline layout and preemptive dialog stack are not being adopted here. Review reproduced cancellation leaking from an expired request into an unrelated queued request, loss of a queued dialog's draft, and keyboard selection moving offscreen. Concurrent request queuing can be addressed separately; sequential select-to-input flows already work without a stack.

Validation: TypeScript, ESLint, and all 891 tests pass. The browser suite passes at 1280px and 390px, including real extension commands for navigation, Escape without agent abort, draft preservation across collapse/expand, countdown updates, and answering a new request after server-side expiry. These checks are now retained in the E2E suite.

PR #683, opened 2026-09-01
fix: let extension widget triggers size to content

### LINKED ISSUE #682: Extension widget trigger strip always reserves 70% width, squeezing the status line
### Summary

When an extension registers a widget, the widget trigger strip always reserves
70% of the status shelf width, even when a single short tab is present. The
extension status line is squeezed into the remaining ~30% and wraps, while most
of the trigger strip is empty space.

### Where

`.extension-status-shelf.has-widgets.has-status .extension-widget-triggers`:

```css
.extension-widget-triggers {
  border-right: 1px solid var(--border);
  flex: 0 70%;
  max-width: 70%;
}
.extension-status-line {
  flex: 1 1 0;
  min-width: 0;
}
```

`flex: 0 70%` sets `flex-grow: 0` with `flex-basis: 70%`, so the strip claims
70% regardless of how many triggers exist or how wide they are. With one tab
labelled `skill-usage`, roughly 55% of the shelf is blank while the status line
wraps onto a second row.

### Expected

The trigger strip should size to its content and only cap at a maximum, letting
the status line use the rest:

```css
.extension-widget-triggers {
  flex: 0 1 auto;
  max-width: 70%;
}
```

`flex-basis: auto` keeps the current behaviour once the tabs actually need the
space (they can still scroll, `overflow: auto hidden` is already set), but a
single tab no longer reserves the full 70%.

### Reproduce

1. Load any extension that calls `ctx.ui.setWidget(key, lines, { placement: "aboveEditor" })`.
2. Also call `ctx.ui.setStatus(key, text)` with a status string of ~60 characters.
3. Observe the status line wrapping while the trigger strip stays mostly empty.

Version: pi-web 0.8.11.

uvforce:
Fixes #682.

`.extension-status-shelf.has-widgets.has-status .extension-widget-triggers`
sets `flex: 0 1 70%`. With `flex-grow: 0` and `flex-basis: 70%`, the trigger
strip claims 70% of the shelf no matter how many triggers exist, so a single
short tab leaves most of that space blank while `.extension-status-line`
is squeezed into the remaining ~30% and wraps onto a second row.

`max-width: 70%` in the same rule already expresses the intended cap, so the
basis only needs to follow the content:

```css
.extension-status-shelf.has-widgets.has-status .extension-widget-triggers {
- flex: 0 1 70%;
+ flex: 0 1 auto;
  max-width: 70%;
  border-right: 1px solid var(--border);
}
```

Behaviour with many triggers is unchanged: the strip still grows up to the
70% cap and keeps scrolling horizontally through the existing
`overflow-x: auto`.

This affects any combination of one widget plus one status line, which is
common in practice — a todo-style widget extension together with an MCP or
model status extension already reproduces it. The rule has no media query,
so on narrow/PWA viewports the status line is reduced to a very small column.

Verified by applying the same declaration to the built stylesheet of a
running pi-web 0.8.11 instance: the trigger strip then sizes to its tab and
the status line renders on one row.

CSS-only change; no component, state, or behaviour changes.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -198,7 +198,7 @@ body {
 }
 
 .extension-status-shelf.has-widgets.has-status .extension-widget-triggers {
-  flex: 0 1 70%;
+  flex: 0 1 auto;
   max-width: 70%;
   border-right: 1px solid var(--border);
 }


agegr:
Thanks for the fix. The same issue has already been resolved by #670, which is merged into main. Review found no defects in this change; Chromium checks across desktop and narrow viewports, single and multiple widgets, and expanded and collapsed panels produced the same layout with this proposal and the fix in #670. Closing as a duplicate of #670.

PR #684, opened 2026-09-01
fix: keep substantial assistant answers visible

### LINKED ISSUE #622: Long assistant text is folded into ProcessDetailsGroup when a trailing tool call follows; UI shows only the last short segment
## Summary

In a single turn, only the **last** assistant message containing any text is rendered standalone. Every earlier assistant message — including one that carries the full main answer — is folded into the collapsed `ProcessDetailsGroup`. If the model writes its answer and then makes one trailing tool call (followed by a short closing line), the UI shows only the closing line; the actual answer is hidden behind "process details".

## Environment

- pi-web 0.8.9 (also reproduced on current `main`)

## Root cause

`components/ChatWindow.tsx`:

1. `findFinalAssistantIndex` (line ~146) scans the turn **backwards** and returns the *last* assistant message with any answer block via `hasFinalAssistantAnswer` — a single non-empty text line qualifies.
2. All other messages in the turn are grouped into `ProcessDetailsGroup` with `defaultExpanded = false` (line ~210).

So for a turn shaped like:

```
user message (anchor)
  ├─ assistant text  ← the real, long answer (tables, details)
  ├─ assistant toolCall + toolResult (e.g. memory write)
  └─ assistant text  ← two-line closing note   ← only this one is visible
```

the long answer is classified as "process" and collapsed, which reads as if pi web "lost" the response.

## Expected behavior

Assistant messages with substantial prose answers should not be treated as foldable process detail. Possible fixes (any one helps):

- Exclude assistant messages whose text blocks exceed a threshold (e.g. > 200 chars, or contain markdown structure such as tables/headings) from the fold group; or
- Pick the *longest* answer-bearing assistant message as the visible final answer instead of the last one; or
- Auto-expand `ProcessDetailsGroup` when it contains answer-bearing text.

## Why it matters

Agent harnesses commonly end a turn with housekeeping tool calls (memory writes, session saves) after the main answer. The current rule makes the main answer invisible by default in exactly those turns.

huiq777:
## Summary

- auto-expand process details when folded assistant messages contain substantial text or images
- keep routine short progress messages collapsed when a final answer is visible
- preserve the existing expanded state when no final answer exists
- add regression coverage for text, image, and fallback behavior

Fixes #622

## Testing

- `npm test` — 848 passed
- `npm run lint`
- `./node_modules/.bin/tsc --noEmit`
- `git diff --check`

### DIFF
--- a/components/ChatWindow.process-details.test.mjs
+++ b/components/ChatWindow.process-details.test.mjs
@@ -4,10 +4,10 @@ import test from "node:test";
 
 const source = await readFile(new URL("./ChatWindow.tsx", import.meta.url), "utf8");
 
-test("expands process details when a completed turn has no final answer", () => {
+test("uses message content to choose the initial process-details state", () => {
   assert.match(source, /const \[expanded, setExpanded\] = useState\(defaultExpanded\)/);
   assert.match(
     source,
-    /<ProcessDetailsGroup[\s\S]*?defaultExpanded=\{!finalAnswerMessage\}/,
+    /<ProcessDetailsGroup[\s\S]*?defaultExpanded=\{shouldExpandProcessDetails\(processMessages, \{ hasFinalAnswer: Boolean\(finalAnswerMessage\) \}\)\}/,
   );
 });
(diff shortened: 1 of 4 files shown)


agegr:
Thanks for working on #622. I'm closing this PR because the proposed default expansion rule is too broad. A hard 200-character cutoff does not reliably distinguish a substantive answer from a progress update, and a single qualifying text block expands the entire process group, including all its tool details.

The usual flow is to finish tool work and then provide the complete final answer. The case where housekeeping tools run after the main answer is worth addressing, but it does not justify changing the default expansion behavior for ordinary long progress messages. For now, we will keep the existing collapse behavior and leave #622 open for a more targeted solution.

PR #694, opened 2026-09-03
fix: use monospace font in Markdown previews

### LINKED ISSUE #638: Markdown Viewer should use monospace font
当前markdown viewer的字体因为不是等宽，导致LLM绘制的ASCII 图形变形

------

Currently, the font used in the Markdown viewer is not monospace, which causes ASCII graphics drawn by the LLM to become distorted.

Radiummm:
## Summary

- use the existing monospace font stack in rendered Markdown file previews
- keep the change scoped to the file viewer so chat typography is unchanged
- add regression coverage for the preview style

## Tests

- `npm test` (845 passed)
- `npm run lint`

Closes #638

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -1048,6 +1048,7 @@ pre, code {
 }
 
 /* File viewer markdown preview — larger headings */
+.markdown-file-preview { font-family: var(--font-mono); }
 .markdown-file-preview h1 { font-size: 1.8em; }
 .markdown-file-preview h2 { font-size: 1.4em; }
 .markdown-file-preview h3 { font-size: 1.15em; }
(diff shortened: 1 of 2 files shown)


agegr:
Plain-text code blocks in Markdown file previews already use the existing `--font-mono` stack on `main`, including fences without a language and those marked `text` or `plaintext` ([CodeBlock](https://github.com/agegr/pi-web/blob/edf574a75209dd4352584f46c3400f8a8b70ff26/components/MermaidBlock.tsx#L280-L293)). ASCII diagrams should be placed in code blocks so their spacing and line breaks are preserved.

This PR changes the font of the entire preview, including ordinary headings and prose. We prefer to keep proportional fonts for that content, so we are closing this PR. Changing the body font also does not address whitespace collapsing outside code blocks. A minimal Markdown example reproducing #638 inside a code block would help us identify and fix the underlying rendering issue.

PR #696, opened 2026-09-03
feat: add read-only plan mode

Radiummm:
## Summary

- add a built-in `/plan` extension and toolbar toggle
- persist plan mode per session and restore the previous tool selection on exit
- expose only known read-only exploration tools while planning
- block direct shell commands and unknown extension tools from bypassing plan mode
- add English, Simplified Chinese, and Traditional Chinese labels

## Tests

- `node --experimental-strip-types --test lib/plan-mode-extension.test.mjs`
- `npm run lint`
- `npm test` (846 passed)


agegr:
The existing read-only tool preset covers the main use case, so we are closing this PR to avoid adding a separate mode and state management. Any stricter read-only restrictions should build on the existing preset.

Issue #628, opened 2026-08-27
Add "Edit from here" and "New session" actions to the first message in the chat


AndSDev:
Add "Edit from here" and "New session" actions to the first message in the chat. This is useful for "Answer with different models" and "Answer with additional info".

agegr:
fixed in 585d56c2b939f16a2551b37e7d57141fb7ad55d4

PR #707, opened 2026-09-04
feat(chat): edit user messages inline

Zhangs-11:
## Summary

- add a pencil action beside Copy for editing a sent user message in place
- replace the message bubble with an inline editor containing Cancel and Send actions
- navigate to the point immediately before the selected message and resend the revised prompt on a new branch of the same session
- support the first user message, text plus image messages, `Escape` to cancel, and `Ctrl`/`Cmd+Enter` to submit
- expose editing only while the session is idle, so an active run must be stopped explicitly first

The session JSONL remains append-only. Older branches stay available through the existing branch navigator, while the active chat view shows only the revised message and its new response. No additional sidebar session is created.

This addresses the editing portion of #628.

## Implementation notes

- `navigate_tree` returns the resulting active leaf so the client can reload the exact parent context before resending
- the context endpoint distinguishes an omitted leaf from an explicit root position, enabling edits to the first message without leaking pagination state from the previous branch
- English, Simplified Chinese, and Traditional Chinese labels are included

## Testing

- `npm test` — 892 tests passed
- `npm run lint`
- real browser QA on the rebased commit: editing and resending the first message hides the old active path and returns a fresh model response in the same session

## Demo

The demo edits the first active user message and reruns it through the real model flow.

Edit and resend a user message inline

Recorded from `Zhangs-11/pi-web@9596cd11e895d44e908fca2d9655f224c10ca93d` on the clean rebased branch, served at `http://127.0.0.1:30159` with `next dev` because repository instructions prohibit `next build` during development. The run used a fresh isolated Pi agent directory, workspace, and headless Chrome profile with the normal application transport; the repository-declared Playwright CDP fallback was used after agent-browser could not connect to its configured CDP endpoint. Real GPT-6-Astra rounds produced both the original and updated answers; no fixtures, mock transport, DOM-event injection, or test-only hooks were used.

### DIFF
--- a/app/api/sessions/[id]/context/route.ts
+++ b/app/api/sessions/[id]/context/route.ts
@@ -9,7 +9,9 @@ export async function GET(
 ) {
   const { id } = await params;
   const url = new URL(req.url);
-  const leafId = url.searchParams.get("leafId") ?? undefined;
+  const leafId = url.searchParams.has("leafId")
+    ? url.searchParams.get("leafId") || null
+    : undefined;
   const deferThinking = url.searchParams.has("deferThinking");
   const deferToolResultImages = url.searchParams.has("deferMedia");
   // `tail` caps the ancestor chain returned (default 50); `before` rewinds the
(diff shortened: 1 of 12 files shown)


agegr:
Thanks for the contribution. The existing "Edit from here" flow already covers editing and resending messages, so we'd prefer to improve that flow (including first-message support) rather than maintain a separate inline editor. Closing this PR for now.

PR #724, opened 2026-09-05
feat(ui): render extension dialog titles with highlighted code fences and a pinned head line

YsDirector:
## Summary

Splits out the dialog-title rendering improvement from the closed #667, per the upstream request ("The dialog-title rendering improvement is independently useful and can be proposed in a separate PR with focused tests").

When an extension raises a blocking dialog, the `title` can be long and can embed a risky command inside a ```sh / ```bash code fence. Today the whole title is rendered as a single truncated line in the header.

This PR:
1. **Pins the head line**: the title is split at its first newline — the one-line head stays in the dialog header (pre-wrap / break-word), the rest moves into the already-scrollable body, so buttons stay visible for long titles.
2. **Highlights code fences**: ```sh / ```bash segments in the body render as red-bordered highlighted code blocks (red tint + left border), flagging risky commands at a glance; other text keeps pre-wrap multi-line rendering.

Pure split logic is extracted to `lib/dialog-title.ts` so it is testable in isolation.

## Changes

- `lib/dialog-title.ts` — new: `splitDialogTitle` (head/rest) and `splitDialogTitleCode` (text/code segment splitting with focused unit tests in `lib/dialog-title.test.mjs`)
- `components/ChatWindow.tsx` — `renderDialogTitle` + head/rest wiring in `ExtensionDialog`

## Tests

- 8 new unit tests for the split/serialization logic
- Full suite: 914/914 passing

### DIFF
--- a/components/ChatWindow.tsx
+++ b/components/ChatWindow.tsx
@@ -3,6 +3,7 @@ import { registerAbortHandler } from "@/hooks/useKeyboardShortcuts";
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState, type ReactNode } from "react";
 import type { AgentMessage, AssistantContentBlock, AssistantMessage, BashExecutionMessage, BlockingExtensionUiRequest, ExtensionUiRequest, SessionInfo, SessionTreeNode, ToolResultMessage, UserMessage } from "@/lib/types";
 import { normalizeCustomPanelLines } from "@/lib/ansi";
+import { splitDialogTitle, splitDialogTitleCode } from "@/lib/dialog-title";
 import { asBracketedPaste, toTerminalKeyData } from "@/lib/terminal-input";
 import { countToolCallBlocks, getAssistantErrorMessage, getDisplayableAssistantBlocks, isMessageGroupAnchor, splitFinalAssistantBlocks } from "@/lib/message-display";
 import { extractTurnWrittenFiles, type WrittenFile } from "@/lib/turn-written-files";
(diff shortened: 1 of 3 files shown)


agegr:
Thanks for splitting this out. I'm going to close it though, because the design goes further than the problem needs and reintroduces the concern from #667 in a softer form.

**What's actually broken**: the expanded dialog header renders `request.title` without `white-space: pre-wrap`, so newlines collapse into spaces. Pi's TUI shows the title verbatim (`new Text(theme.bold(title))`), newlines included, and `select`/`input` have no separate `message` field, so extensions legitimately put multi-line text there. That is a one-line CSS mismatch with upstream and is being fixed on `main` as exactly that: `pre-wrap` + `overflow-wrap: anywhere` on the title, plus a max height on the header so a long title can't push the options and buttons out of view.

**Why not this PR**:
- It gives `title` semantics Pi doesn't have: "first line is the header, the rest is body, and a ```sh fence means a risky command". The TUI renders none of that, so extensions relying on it would look different in Pi Web than in Pi. Same category as the `__piBatchAsk__` envelope in #667.
- Since #699 (f4a700d), `confirm.message` and `select` options already go through `MarkdownBody`. Code that needs highlighting belongs in `message`, rendered by the same markdown pipeline as the chat, not a second sh/bash-only fence parser in `lib/dialog-title.ts`.
- The regex `/```(?:sh|bash)?\s*\n?/` makes the language optional, so every fence matches and the label leaks into the code text (```python → highlighted block starting with `python`). Unbalanced fences turn the trailing text into code.
- The last test reimplements the mapping inline instead of exercising `renderDialogTitle` in `ChatWindow.tsx`, so it doesn't guard the component.
- Hardcoded `#ef4444` colors bypass the theme variables, and "red = dangerous" is a policy call a generic dialog renderer shouldn't make.

Appreciate the report; the newline fix is the part that was real.

PR #725, opened 2026-09-05
feat(ui): tool calls and thinking enhancement

Eyre-S:
**Adds icons for all internal tools and mcp_ tools**:

圖片

**read/write/edit syntax highlight, along with better diff view**

| 圖片  |   圖片 |  圖片  |
|------|-----|----|

**switchable diff view wrapping**

|   Wrap (default) | No wrap |
|---|---|
|  圖片 |  圖片 |

**Bash command syntax highlight**

圖片

**Pretty tool input parameters**

圖片

圖片

**Two different type of thinking area**

|  | Card(Default) | Minimal |
|---|---------|--------|
| Collapsed | 圖片 | 圖片 |
| Expanded | 圖片 | 圖片 |

---

Does this pr too large? Should I split it to some small PRs?


agegr:
Thanks for the work here — the tool icons and the read/write/bash highlighting look good in the screenshots.

To answer your question: yes, this is too large to review and land as one PR. It bundles six independent changes (tool icons, file-result syntax highlighting, split-diff word highlighting, bash highlighting, JSON param list, and two new Settings toggles) plus two new dependencies. Please split it into focused PRs so each can be reviewed and reverted on its own. Suggested order:

1. `ToolIcon` + `lib/tool-names.ts` additions (small, self-contained).
2. Bash command highlighting (`BashCommandView`).
3. Read/write result rendering (`FileToolResult`, `lib/file-language.ts`, the `EXT_TO_LANGUAGE` additions).
4. Diff improvements (`lib/word-diff.ts`, `lib/diff-highlight.ts`).
5. `JsonParamList` — this one is 391 lines for a param view; I'd like to see it justified on its own.

Two things I'd like dropped rather than split out:

- The **thinking style** and **diff wrap** Settings toggles. pi-web tries to stay a thin wrapper around pi and avoid growing the settings surface; a single good default is preferred over a preference. Pick the one you think is best and ship that.
- Reading `localStorage` on every `ThinkingBlock` render (`getPreferredThinkingStyle()` in the render path) — if a preference does stay, it needs to go through a hook like the existing `useChatAppearance`.

Also: `package.json` adds `diff` and `refractor` but `package-lock.json` is not updated, so `npm ci` fails on this branch. Each split PR that adds a dependency should include the lock change.

Closing this umbrella PR is up to you; I'll review the split PRs as they come in.

Eyre-S:
Thanks for your review.

I will split the works to several PRs these days, and make some changes based on your review.

PR #732, opened 2026-09-06
feat(worktree): fetch origin before add and raise git timeout for large repos

dingxiaobo:
Three worktree fixes:

- Run `git fetch origin` (60s) before resolving the branch so worktrees start from current remote code. Non-fatal — falls back to local refs.
- Make `git()` timeout configurable (default 10s unchanged): fetch → 60s, `worktree add` → 5min for repos with 30k+ files.
- When the branch doesn't exist locally, create it from the remote tip (`refs/remotes/origin/<branch>` populated by the fetch) instead of stale HEAD; fall back to HEAD when the remote ref is absent.

### DIFF
--- a/lib/worktree.ts
+++ b/lib/worktree.ts
@@ -49,9 +49,9 @@ export function invalidateProjectCache(): void {
   globalThis.__piProjectCache?.clear();
 }
 
-async function git(cwd: string, args: string[]): Promise<string> {
+async function git(cwd: string, args: string[], timeoutMs = 10_000): Promise<string> {
   const { stdout } = await execFileAsync("git", ["-C", cwd, ...args], {
-    timeout: 10_000,
+    timeout: timeoutMs,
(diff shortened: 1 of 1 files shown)


agegr (approved):
Thanks. The raised timeouts (5 min for `worktree add`, configurable `git()` timeout) and starting a new branch from `refs/remotes/origin/<branch>` when it exists are both good.

I pushed one change to the branch (b/c the fix is small): removed the implicit `git fetch origin` before every `worktree add`. Creating a worktree should stay a local, offline-safe operation — an unconditional fetch adds network latency (up to 60 s on a bad connection) to every add, including when the branch already exists locally, and it silently fetches every ref on large repos. If the user wants the remote tip they can fetch first; the remote-tracking fallback you added then picks it up.

Verified: tsc clean on top of `main`, `lib/worktree.test.mjs` passes.

PR #733, opened 2026-09-06
feat: add extension widget font size setting

dingxiaobo:
Adds a configurable font size for extension widget content, mirroring the existing chat font size control.

dxb_screen_2026-09-06_14-04-02

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -423,7 +423,7 @@ button.extension-widget-trigger:focus-visible {
   padding: 3px 12px 8px;
   color: var(--text-muted);
   font-family: var(--font-mono);
-  font-size: 14px;
+  font-size: var(--extension-widget-font-size, 14px);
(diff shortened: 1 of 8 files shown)


agegr (approved):
Thanks — the underlying problem is real: extension widget text was pinned at 14px and ignored the chat font size.

I pushed a change to the branch that solves it without a new setting. pi-web tries to keep the Settings surface small, and a second slider for one sub-region of the chat is more knobs than the problem needs. The widget body is rendered inside `.chat-content`, so it now follows the existing chat font size:

```css
.extension-widget-content {
  font-size: calc(14px + var(--chat-font-size-offset, 0px));
}
```

The rest of the PR (new hook state, storage key, i18n, settings row, tests) is reverted; the net diff is one CSS line. Approving in this form.

PR #735, opened 2026-09-06
feat: preview image attachments in the chat composer

wyx20021207-create:
## Summary
- Reuse the existing ImagePreview modal for attachments in the chat composer, before sending.
- Preserve thumbnail sizing and keep the remove button separate from the preview trigger.
- Reuse keyboard activation, Escape/backdrop dismissal, and focus restoration from the shared component.

## Testing
- Shared image-preview regression tests pass.
- `npm run lint` and TypeScript `tsc --noEmit` pass in the development checkout.
- Ran `npm test` on Windows and compared it with an untouched upstream/main checkout at 0d1df12: both have the same 14 failing tests (including Windows symlink/shell limitations and existing component assertions). No additional failing tests were observed.
- Production build and browser E2E were not run locally.

## Manual verification
1. Paste, drag, or attach an image in the composer.
2. Click the thumbnail to open the modal; close with Escape, backdrop, or the close button.
3. Use the remove button and confirm it removes the attachment without opening the modal.

This PR is independent of the local-path preview feature.

### DIFF
--- a/components/ChatInput.tsx
+++ b/components/ChatInput.tsx
@@ -23,6 +23,7 @@ import {
   type AtQueryMatch, type FileIndexEntry,
 } from "@/lib/file-fuzzy";
 import { FolderIcon, getFileIcon } from "./FileIcons";
+import { ImagePreview } from "./ImagePreview";
 import { useIsMobile } from "@/hooks/useIsMobile";
 import { useI18n } from "@/hooks/useI18n";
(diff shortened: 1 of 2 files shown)


agegr (approved):
Small, reuses the existing `ImagePreview` dialog, and the remove button stays outside the trigger so it does not open the modal. Verified on top of current `main`: tsc clean, `ImagePreview`/`ChatInput` tests pass. Thanks.

PR #736, opened 2026-09-06
feat: preview verified local files and directories from chat

wyx20021207-create:
## Summary
- Recognize local path references in Markdown text and inline code, without rewriting fenced code, math, or existing links.
- Validate candidates using the existing authorized file metadata API before making them clickable. Nonexistent/inaccessible paths retain their original text.
- Keep the original path label and neutral code-style background; show the resolved full path on hover and open it in the right-hand panel.
- Support directory metadata and a right-panel directory listing, with navigation into child directories and files.
- Recheck before opening, periodically refresh validation, deduplicate concurrent checks, and cap metadata request concurrency.
- Keep the existing allowed-root and realpath checks; do not broaden filesystem access.

## Testing
- Markdown rendering, path validation, directory dispatch, and existing file-viewer regression tests pass.
- `npm run lint` and TypeScript `tsc --noEmit` pass in the development checkout.
- Ran `npm test` on Windows and compared it with an untouched upstream/main checkout at 0d1df12: both have the same 14 failing tests (including Windows symlink/shell limitations and existing component assertions). No additional failing tests were observed.
- Production build and browser E2E were not run locally.

## Manual verification
1. Include an existing relative path or Windows absolute path in a chat response.
2. Verify the original label remains visible, the full path appears on hover, and clicking opens the right panel.
3. Include a nonexistent path or slash-separated function names and verify they remain plain text/code.
4. Click an existing directory and navigate through its entries in the right panel.

This PR is independent of the composer image-preview feature.

### DIFF
--- a/app/api/files/[...path]/route.ts
+++ b/app/api/files/[...path]/route.ts
@@ -496,6 +496,9 @@ export async function GET(
     }
 
     if (type === "meta") {
+      if (stat?.isDirectory()) {
+        return NextResponse.json({ isDirectory: true });
+      }
       if (!stat?.isFile()) {
         return NextResponse.json({ error: "Not a file" }, { status: 400 });
       }
--- a/app/globals.css
+++ b/app/globals.css
@@ -533,6 +533,43 @@ pre, code {
   font-size: 0.92em;
   box-shadow: inset 0 0 0 1px color-mix(in srgb, var(--border) 70%, transparent);
 }
+.markdown-body a.markdown-local-file-link,
+.markdown-body a.markdown-local-file-link:hover {
+  color: var(--text);
+  text-decoration: none;
+  cursor: pointer;
+  background: var(--bg-subtle);
+  border-radius: 5px;
+  padding: 1px 5px;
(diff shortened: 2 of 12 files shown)


agegr:
Thanks for the careful implementation (the conservative regex, the allowed-root checks, and the tests are all appreciated). I'm going to close this on product scope rather than code quality.

pi-web already opens local files from explicit Markdown links in assistant output (`resolveLocalFileHref` in `MarkdownBody`), and that is the level of magic I want to keep. This PR adds:

- a heuristic pass over every text/inline-code node in every message to guess which tokens are paths,
- a metadata request to `/api/files` per candidate, with dedup, concurrency caps, and periodic re-validation,
- a second viewer (`DirectoryViewer`) with its own navigation.

That is a lot of moving parts (and background request traffic) for turning bare paths into links, and the heuristic will always have false positives/negatives to chase. pi-web is meant to stay a thin wrapper around pi, so I'd rather leave this to the model emitting real links when it wants something clickable.

PR #737, opened 2026-09-06
fix: serialize extension service loading

lyx2145181:
## Summary

- Serialize Pi SDK service construction so concurrent model discovery and session startup cannot race extension factories.
- Route Pi Web model discovery and RPC session startup through the shared lifecycle-aware service loader.
- Restore transient extension state after service-only model discovery so extensions such as `pi-chrome` do not consume a singleton without receiving `session_shutdown`.
- Keep the change focused: personal UI customizations, LAN defaults, planning documents, and unresolved background-race changes are not included.

## Why

Pi Web can evaluate extension factories from multiple entry points. A service-only model discovery request may overlap with real session startup, and transient extension loading has no session lifecycle callback to release resources. Serializing these loads and cleaning up transient state prevents missing extension tools and makes startup behavior deterministic.

## Validation

- `npm test` — 954 passed
- `npm run lint`

## Scope

This PR is based on the latest `upstream/main` (`v0.9.0`) and contains only the focused, independently validated changes from the `upstream/agent-lifecycle` branch.

### DIFF
--- a/app/api/models/route.ts
+++ b/app/api/models/route.ts
@@ -1,6 +1,6 @@
 import { stat } from "fs/promises";
 import { resolve } from "path";
-import { createAgentSessionServices, getAgentDir, type SettingsManager } from "@earendil-works/pi-coding-agent";
+import { getAgentDir, type SettingsManager } from "@earendil-works/pi-coding-agent";
 import { getSupportedThinkingLevels } from "@earendil-works/pi-ai";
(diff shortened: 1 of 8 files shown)


lyx2145181:
Closing this narrowly scoped draft so the changes can be reorganized into complete, reviewable performance, reliability, and UI pull requests.

PR #743, opened 2026-09-06
Keep user-opened tools expanded across streaming updates

draxxris:
Currently user can open tools/thinking blocks, but when a new streaming event comes in, the blocks are closed.

This PR keeps them open.

### DIFF
--- a/components/ChatWindow.tsx
+++ b/components/ChatWindow.tsx
@@ -270,6 +270,20 @@ export function ChatWindow({ session, searchTarget, onSearchTargetHandled, initi
     return position && !position.atBottom ? position : null;
   });
   const [restoreAnchorReady, setRestoreAnchorReady] = useState(false);
+  // Lifted expanded state for tool calls — survives streaming re-renders and
+  // stream → history promotion (keyed by toolCallId, fallback to stream/entry index).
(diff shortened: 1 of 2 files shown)


agegr (approved):
Confirmed the bug: a tool card opened while streaming gets remounted twice — once when `message_end` moves the message from `streamState` into `messages` (no `entryId` yet, so the key is the index), and again when `entryIds` arrive from the session file — and each remount reset `useState(false)`.

The branch had conflicted with `main`, so I merged `main` and replaced the implementation with a lighter one (same behavior, credit to you for the report and the toolCallId keying):

- `lib/tool-call-expansion.ts` — a module-level `Set<toolCallId>`; `ToolCallBlock` seeds its `useState` from it and writes back on toggle.
- No props threaded through `ChatWindow` → `MessageView` → `AssistantMessageView` → `BlockView`, and no new `Set` identity in the `MessageView` memo comparator — with the lifted-state version every toggle re-rendered every message in the list.

Verified: tsc clean, eslint clean, `MessageView` tests plus a new `lib/tool-call-expansion.test.mjs` pass. Thinking blocks already have their own expansion preference, so they are left as they were.

PR #744, opened 2026-09-06
Render apply_patch tool calls (GPT) as split diffs

draxxris:
Support rendering apply_patch tool calls (OpenAI's version of edit) as split diffs, like the built-in edit tool. Supports at least: https://github.com/code-yeongyu/pi-apply-patch


agegr (approved):
Looks good — self-contained parser with tests, and it reuses the existing `SplitDiffFile` model so the built-in edit view and `apply_patch` render the same way. Supporting a popular pi extension's tool output is in scope.

Two things I did on the branch:

- Merged current `main` (clean).
- Fixed a small regression: the input-args condition dropped `isStreamingInput`, so an `edit` call with its arguments still streaming showed nothing when expanded. Restored `(isStreamingInput || !isEditTool) && !patchFiles`.

Verified: tsc, eslint, and the `apply-patch` / `turn-written-files` / `MessageView` / `TurnWrittenFiles` tests pass.

PR #747, opened 2026-09-07
feat(chat): undo and edit an aborted first turn

### LINKED ISSUE #746: Support undo/edit for the first message after stopping generation
## Problem

Currently, the first user message in a conversation cannot be edited or undone.

For later messages, pi-web allows users to edit them through the conversation branching mechanism. However, the first message is a special case: there is no previous node to branch from, so it cannot be edited in the same way.

This is inconvenient when I start a new conversation, send the first prompt, and immediately realize that something is wrong — for example:

- a typo;
- an incorrect command or path;
- missing context;
- the wrong file/reference;
- or simply a prompt that I did not intend to send.

In this situation, I usually notice the mistake almost immediately and click **Stop**.

At that point, the assistant may already have generated a small amount of text, but generation has usually only just started. In most cases, the model has not reached the stage of calling tools or performing any external actions yet.

Currently, the practical workaround is to abandon the conversation and create a new one.

## Desired behavior

After stopping generation for the first turn, it would be useful to provide an **Edit**, **Undo**, or **Edit and resend** action.

A possible workflow would be:

1. Start a new conversation.
2. Send the first user message.
3. Immediately notice that the message is incorrect.
4. Click **Stop**.
5. Click **Edit / Undo**.
6. Discard the first user message and any partial assistant response generated from it.
7. Restore the original user message to the input box.
8. Edit it and send it again.

The conversation would then continue as if the mistaken first turn had never been sent.

## Why discarding the stopped turn is useful

This is slightly different from normal conversation branching.

When I intentionally edit an older message, preserving the original branch makes sense because I may want to keep both conversation paths.

But when I notice an error immediately after sending the very first message, I usually do not want to preserve that mistaken turn at all.

Because I typically stop generation immediately, the assistant has usually produced only a small partial response and has not started any tool calls. In that common case, discarding the unfinished first turn from the conversation history should be a natural and lightweight operation.

If a side-effecting tool call has already started or completed, pi-web could disable this behavior, show a warning, or only remove the conversation history without implying that external side effects have been reverted.

## Expected semantics

Conceptually, these two operations serve different purposes:

- **Branch / Fork** → preserve the existing history and explore an alternative path.
- **Undo / Edit and resend** → discard the mistaken latest turn and replace it.

For the first message specifically, undoing it could simply return the conversation to its initial empty state instead of requiring the user to create another conversation.

wyx20021207-create:
Fixes #746

## Summary

Add an “Undo and edit” action for a mistaken first prompt after generation has been stopped.

- Restore the original text and image attachments to the existing composer.
- Permanently discard the first user message and its aborted assistant reply.
- Keep the same session, working directory, and initial settings.
- Do not create a new session or preserve a history branch.
- Run directly on click, without a browser confirmation dialog.
- Reuse the existing composer rather than introducing an inline message editor.

## Scope and safety

This is limited to an original, unbranched first turn with one user message and one aborted assistant reply containing only text/thinking.

- Reject completed turns, later turns, tool calls, and dependent history.
- Reject running sessions, queued prompts, and active background work.
- Refuse to overwrite an existing composer draft.
- Restore the input before the destructive request.
- Shut down the session wrapper before atomically replacing the file.
- Reject the operation if the session changes during shutdown.
- Preserve initial session metadata and configuration.

This operation does not undo filesystem changes or external side effects.

## Testing

- 6 targeted tests pass, including a real SessionManager undo/reopen/resend round-trip.
- TypeScript checks and lint pass.
- Full test suite on the standalone PR branch (Windows): 940 passed, 14 failed, 5 skipped. The same 14 failures were previously reproduced on untouched upstream.

## Documentation

Document eligibility, destructive semantics, concurrency safeguards, and test coverage.


agegr:
Thanks for the contribution. The core first-message edit/new-session workflow is now covered by #628, so I'm closing this to avoid overlapping implementations.

PR #754, opened 2026-09-07
feat: materialize pasted/dropped chat images to server temp file paths (#635)

### LINKED ISSUE #635: Feature: Materialize pasted/dropped images to server-side temp file paths (TUI parity)
## Problem
In the pi TUI, `ctrl+V` writes the clipboard image to a real temp file (e.g. `/tmp/pi-clipboard-*.png`) and the agent receives a **file path**. This gives the agent (and the user) an addressable, re-attachable artifact on disk.

In pi-web, images are attached (paste/drag-drop/📎) as **inline base64** in the message content (`{ type:"image", source:{ type:"base64", data, media_type } }`). No file is ever written server-side. Consequences:
- No `/tmp` file exists for the image — the TUI-visible `pi-clipboard-*.png` behavior is absent.
- The agent cannot `read`/reference the image by path; it only has the inline bytes.
- There is no re-attachable on-disk artifact (e.g. to reuse the same screenshot in a later prompt).

## Desired behavior
When the user pastes/drops/attaches an image in pi-web, the **server** should write the received bytes to a temp file (e.g. `$TMPDIR/pi-web-clipboard-<uuid>.<ext>`) and pass the **file path** to the agent — matching TUI semantics. The UI can keep showing its inline preview; the server just materializes a path for the agent layer.

## Why server-side
The browser is sandboxed and cannot write to the server filesystem. The server already receives the base64 bytes (it validates them), so writing them to a temp file before handing to the agent is a small, localized change.

## Design notes (from maintainer review)
- **File extension** must come from the image `mimeType` (the TUI does `extensionForImageMimeType`); don't hardcode `.png`.
- **Naming**: `pi-web-clipboard-*` is deliberately different from the TUI's `pi-clipboard-*` to avoid collisions/confusion if both run on the same host.
- **Permissions**: write temp files with `0600` (the bytes are user data).
- **Cleanup**: add a GC story — temp files are ephemeral; the TUI's own file is single-shot and not stable across sessions, so "re-attachable" should be scoped to the current session/process, not treated as durable storage.
- **Scope**: the AC "path stable across the session" conflicts with temp-file semantics; reattach-across-sessions needs a separate persistence design and should be out of scope here.

## Acceptance criteria
- [ ] Pasting/dropping an image in pi-web creates a temp file on the server (visible in `$TMPDIR`), not just inline base64.
- [ ] The agent receives a file path (or the file is made available to the agent's `read`/file tools) for the attached image.
- [ ] Temp files are written with `0600` perms, extension derived from `mimeType`, and a GC/cleanup mechanism prevents orphan accumulation.
- [ ] Existing behavior (inline base64 for vision models) is preserved where it doesn't conflict.

## Related
- #584 — different but adjacent: images are silently dropped for text-only models (no warning). This request is about **materializing to a path**, orthogonal to the vision/no-vision UX.
- #440 — umbrella.

## Environment
- pi-web 0.8.11 (current)
- pi / @earendil-works/pi-coding-agent 0.84.3

Shadowzzh:
Closes #635

## Summary

When a user pastes, drops, or picks an image in pi-web, the server now writes the received bytes to private temporary files and appends their absolute paths (numbered in attachment order) to the agent's message, matching the TUI's clipboard semantics. Inline image inputs are preserved unchanged for vision models, so each path is an additional, addressable, re-attachable artifact for local file/image tools.

## Implementation

- `lib/image-materialization.ts` (new): `ImageAttachmentStore` writes each batch atomically under `$TMPDIR/pi-web-clipboard-v1-<pid>-<rand>/batch-<rand>/`, with per-image UUID filenames and extensions derived from the declared MIME type (client-supplied names/paths are never used). Directories use `0700` and files `0600` on POSIX. Cleanup is layered: an hourly GC expires the process's files after 24 h of inactivity (refreshed by new submissions, deferred while any session is busy or has queued messages), orphan directories older than 24 h whose owning PID no longer exists are collected on later uploads (skipping live owners, other users' directories and symlinks), and normal process exit removes everything.
- `lib/rpc-manager.ts`: `prompt` / `steer` / `follow_up` commands materialize their images before dispatch and append the path annotation to the message. A batch is rolled back only when the submission is rejected before preflight acceptance; files survive provider failures after acceptance so follow-up work can reuse the referenced paths. Text-only messages allocate nothing.
- Tests & docs: storage unit tests, RPC integration tests (including real `read`-tool access), a browser E2E exercising picker/paste/drop on desktop and mobile widths, and a design doc (`docs/image-attachments.md`).

## Acceptance criteria

- [x] Pasting/dropping an image in pi-web creates a temp file on the server (visible in `$TMPDIR`), not just inline base64.
- [x] The agent receives a file path (or the file is made available to the agent's `read`/file tools) for the attached image.
- [x] Temp files are written with `0600` perms, extension derived from `mimeType`, and a GC/cleanup mechanism prevents orphan accumulation.
- [x] Existing behavior (inline base64 for vision models) is preserved where it doesn't conflict.

## Design notes

- Naming uses `pi-web-clipboard-*`, deliberately distinct from the TUI's `pi-clipboard-*`, to avoid collisions when both run on the same host.
- Lifecycle is session/process scoped and ephemeral: paths can expire on restart or OS temporary-directory cleanup; historical messages keep their inline image data but old paths are not recreated. Cross-session path permanence is intentionally out of scope here (per the issue's design notes).

## Verification

- Based on `ce18006` (post-v0.9.0, pi-coding-agent 0.85.1) with no drift from `main`.
- `npx tsc --noEmit` clean; `npm run lint` clean; `npm test` 969/969 pass (16 new tests covering byte preservation, private permissions, batch isolation/rollback, GC & orphan collection, and the RPC preflight accept/reject lifecycle).
- Browser E2E (`e2e/image-attachments.mjs`) runs in the CI e2e job and verifies readable server files at the SDK extension boundary without contacting a model provider.

### DIFF
--- a/docs/image-attachments.md
+++ b/docs/image-attachments.md
@@ -1,0 +1,65 @@
+# Temporary files for chat images
+
+Pasted, dropped, and selected chat images are still sent as inline image inputs
+and shown in the composer and conversation. Pi Web also saves the received image
+bytes to temporary files on the server and appends their absolute paths, numbered
+in attachment order, to the agent's message. Local file and image tools can use
+these paths without asking the user to locate or save the image again (#635).
+
+Files are created before the prompt is submitted. A failed write rejects the
+submission and removes that batch's partial files. A rejected prompt preflight
+also removes its batch; a provider failure after acceptance preserves the files
+for follow-up work. Normal prompts, streaming prompts, steering and follow-ups
+use the same storage path.
+
+## Storage and lifetime
+
+- Files live below the OS temporary directory (`os.tmpdir()`, normally `$TMPDIR`
+  on Unix) in a private `pi-web-clipboard-v1-<pid>-<random>/batch-<random>` directory.
+- File names are `pi-web-clipboard-<uuid>.<extension>`. The extension comes from
+  the declared image MIME type; client-supplied names and paths are never used.
+- Directories use mode `0700` and files use `0600` on POSIX systems. Windows
+  access remains governed by the user's temporary-directory ACLs.
+- The existing limits of 10 images and 10 MiB per image remain. Malformed Base64
+  and MIME types without a supported file extension are rejected. This is not a
+  full image decoder: existing image decoding remains the responsibility of the
+  browser, model provider and image tools.
+- Files belong to the server process, so releasing an idle session wrapper,
+  reloading its tools, or forking a session does not immediately delete them.
+- The process checks hourly and removes its temporary files after 24 hours of
+  inactivity. New prompt/steering/follow-up submissions refresh that period;
+  active agent work and queued messages defer expiry for all images in the
+  process. These files are not a durable storage API or a disk quota mechanism.
+- Normal process exit removes its files. On the next image upload and hourly
+  thereafter, directories older than 24 hours whose owning process no longer
+  exists are collected. Live or unverifiable owners, other users' directories
+  on POSIX, unrelated names and symlinks are skipped. PID reuse conservatively
+  delays collection. Cleanup errors are logged by the server.
+
+Paths can expire or disappear on restart or OS temporary-directory cleanup.
+Historical messages keep their inline image data, but their old paths are not
+automatically recreated. Reattach the image when a fresh path is needed. No
+cross-session or cross-process path permanence is promised. The file browser's
+allowed roots are not broadened to include the whole temporary directory.
+
+## Image fidelity
+
+The temporary file contains exactly the bytes received by the server. The
+existing browser compression still applies: images larger than 1 MiB, except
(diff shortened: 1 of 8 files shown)


agegr:
Thanks for the thorough implementation and test coverage.

We have closed #635 as not planned after reviewing the product behavior and existing file-upload workflow. Automatically materializing every image would add hidden filesystem side effects, persist expiring absolute paths in conversation history, and require cleanup machinery in the default attachment path. The narrower tool-oriented use case does not justify changing the default image behavior at this time.

This is a scope decision rather than a concern about the care put into the PR. If repeated user demand emerges, a smaller explicit "save to project and mention" workflow that reuses the existing upload path would be a better follow-up.

Closing this PR with the underlying issue.

PR #761, opened 2026-09-07
feat: bottom-dock extension dialogs, declutter titles, unlock full extension UIs

luckrnx09:
## What

Three interaction fixes for extension UI requests (select/confirm/input/editor + custom panels), reported in #679 and related pain points:

1. **Bottom-docked cards instead of centered modals** — `ExtensionDialog` and `ExtensionCustomPanel` now anchor above the composer. The chat stays scrollable while a request is open, so users can re-read context before answering (#679).
2. **Multi-paragraph titles split into header + body** — hosts/extensions that fold context into the title (previews, instructions, `[header]` prefixes) now render the first paragraph as the header and the rest as scrollable body text, instead of cramming everything into the title row.
3. **Attention sound debounced across chained dialogs** — a select followed by an input (the "type your own answer" flow) is one interaction and now rings once, not twice.
4. **Advertise `mode: "tui"` to extensions** — pi-web runs the agent in-process and its `uiContext.custom()` is fully implemented (terminal-rendered panel), but extensions were told `"rpc"`, so extensions with rich TUI components (e.g. tabbed multi-question questionnaires with checkboxes) degraded to sequential select/input dialogs. The label now matches actual capabilities, unlocking their native UIs.

## Verification

- `npm test` (956 pass), `eslint`, `tsc --noEmit` clean
- Source-assertion tests added for docking, sound debounce, and title splitting
- Playwright geometry check: dialog bottom sits above the composer, chat scrolls while open
- Real-extension end-to-end check with a questionnaire extension: full multi-select checkbox questionnaire renders and submits through the custom panel

Collapsed-pill behavior, keyboard navigation, draft preservation, and Esc semantics are unchanged (covered by existing e2e fixture).

### DIFF
--- a/components/ChatWindow.extension-request.test.mjs
+++ b/components/ChatWindow.extension-request.test.mjs
@@ -33,3 +33,21 @@ test("resets collapse state when a new extension request arrives", () => {
   assert.match(source, /<ExtensionCustomPanel key=\{extensionCustomUi.id\}/);
   assert.match(customSource, /if \(!collapsed\) inputRef.current\?\.focus\(\);\s*}, \[collapsed\]\)/);
 });
+
+test("docks extension overlays to the bottom so the chat stays scrollable", () => {
+  assert.match(dialogSource, /alignItems: collapsed \? "flex-start" : "flex-end"/);
+  assert.match(customSource, /alignItems: collapsed \? "flex-start" : "flex-end"/);
+  assert.doesNotMatch(dialogSource, /alignItems: collapsed \? "flex-start" : "center"/);
+  assert.doesNotMatch(customSource, /alignItems: collapsed \? "flex-start" : "center"/);
+});
+
+test("debounces the completion sound across chained dialogs", () => {
+  assert.match(source, /EXTENSION_DIALOG_SOUND_MIN_GAP_MS = \d+/);
+  assert.match(source, /now - extensionDialogLastSoundAtRef\.current < EXTENSION_DIALOG_SOUND_MIN_GAP_MS/);
+});
+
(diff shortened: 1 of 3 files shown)


luckrnx09:
@agegr Adapt and extend in the web UI (especially custom terminal UI) is really tricky. 

How about just using the custom UI from the Terminal? Although its interaction feels quite abrupt compared to web components, it would ensure that the extension's functionality matches TUI without introducing compatibility issues. 

This PR is just my initial idea and hasn't been fully tested yet.
If you're interested, let me know what you think so I can decide whether to move forward with it or not.

agegr:
Thanks for sketching this out and for asking before investing more. Closing it, with reasons per item:

1. **Bottom-docked dialogs** — #679 was already addressed in #612 (dialogs can be collapsed, the chat scrolls behind them, and input/stop are not blocked). I don't want to re-layout the dialogs again on top of that.
2. **Multi-paragraph titles** — landed on `main` in fcd94bf ("preserve newlines in extension dialog titles like pi's TUI"), which follows pi's own TUI rendering rather than inventing a header/body split.
3. **Debounced attention sound** — reasonable on its own; if you want to open a tiny PR that only does that, I'll review it.
4. **Advertising `mode: "tui"`** — this one I'd decline outright. pi-web's `uiContext.custom()` is not a real terminal, and extensions branch on `mode` to decide whether they can rely on TUI-only behavior (focus, key handling, layout). Telling them we are a TUI trades a known, honest degradation for unknown breakage. If a rich extension UI matters, the right fix is in the extension's `rpc` fallback (tracked upstream).

The branch is also conflicting with `main` now. Appreciate the thought that went into it.

PR #776, opened 2026-09-08
Let a dedicated model name sessions

uvforce:
## Problem

Session titles are generated with the session's own model. Naming a conversation therefore spends a request on whatever large model happens to be selected, and `buildSessionTitleAgentOptions` copies that session's `thinkingLevel`, so a one-line title can be produced with a full reasoning budget.

## Change

- **General settings → Session title** selects the model used for naming. The default, *Use the session's own model*, keeps today's behaviour.
- The choice is stored as `sessionTitleModel` in `settings.json`, using the same locked read/write pattern as the existing PowerShell tool preference, and served by `/api/session-title/settings` following the shape of `/api/tools/settings`.
- `generateSessionTitle()` accepts a `provider/model-id` override and resolves it through the session's own `modelRuntime`, so only configured models can be used.
- Thinking is reduced to the cheapest level each model accepts. Gemini gets `low` because it rejects `minimal` and its SDK reports that level whenever thinking is disabled.

Tool shadowing and message sanitization are untouched, so naming still cannot execute tools or mutate the source session.

## Tests

- `lib/session-title-settings.test.mjs` — defaults, persistence, removal of the key when inheriting, `0600` permissions.
- `app/api/session-title/settings/route.test.mjs` — GET/PUT round trip and payload validation.
- `lib/session-title.test.mjs` — model override and thinking level resolution.
- Full `npm test` (956), `tsc --noEmit`, `eslint`, and `next build` pass.
- Verified in a production build with a real browser: the section renders in General, the stored model is shown when the catalogue is unavailable, and switching back to the default removes the key from `settings.json`.

## Notes

The model catalogue is project-scoped, so the dropdown lists models only when a project is open; the stored value is still displayed without one.


uvforce:
Addressed all three review points in cefd1e4:

- A configured model that the runtime cannot resolve now fails with `Session title model not found` instead of silently naming with the session model. The runtime is refreshed first (`refresh({ allowNetwork: false })`, as `set_model` does), so a model added while this session's runtime was already alive is still found.
- `resolveTitleThinkingLevel` takes the first entry of `getSupportedThinkingLevels()`, which is already in ascending cost order. A model excluding `off` gets `minimal`; one exposing only `medium` and above gets `medium`. Gemini still skips `minimal`.
- The model catalogue reloads whenever General becomes the active section, so a model added under Models is selectable without closing the dialog.

Tests cover the refresh-then-reject path, both level cases, and the reload wiring.

agegr:
Thanks for the work. After review, I don’t think this has enough value to merge as-is, so I’m closing it for now.

agegr:
One additional consideration: reusing the same model can preserve prompt-cache hits for the full conversation history. Switching to a dedicated model may lose that cache, so using a smaller or cheaper model does not necessarily reduce the total cost.

PR #777, opened 2026-09-08
Show the reasoning level of the running turn

uvforce:
## Problem

The reasoning level control is hidden while a turn is streaming, because the level cannot be changed mid-turn. That also removes the answer to the question that matters most at exactly that moment: which budget is this turn spending. A long-running turn gives no indication of whether it is on `auto`, `low`, or `xhigh`, and the level is only visible again once the turn is over.

## Change

While streaming, the control row shows the active level as a read-only label in the same slot the dropdown normally occupies. Nothing else about the streaming controls changes.

It is deliberately not a button: the level cannot be changed until the turn ends, so a clickable control there would invite clicks that do nothing. The label reuses `thinkingDisplayLabel`, the existing `ThinkingIcon`, and the same padding, height, and mobile collapse rule as the editable control, so the row keeps its spacing next to Stop.

## Tests

- `components/ChatInput.streaming-thinking.test.mjs` — the streaming branch renders the level without any button or click handler, and shares padding, height, and the mobile collapse rule with the dropdown.
- Verified in a production build with a real browser: during a streaming turn the level appears next to Stop, and the editable dropdown returns unchanged once the turn ends.

### DIFF
--- a/components/ChatInput.streaming-thinking.test.mjs
+++ b/components/ChatInput.streaming-thinking.test.mjs
@@ -24,3 +24,12 @@ test("keeps the read-only level aligned with the neighbouring controls", () => {
   // The label collapses to the icon on mobile exactly like the editable control.
   assert.match(readOnly, /\(!isMobile \|\| controlsMenuOpen\) &&/);
 });
+
+test("shows the level the runtime actually applies, not the selector placeholder", async () => {
+  const session = await readFile(new URL("../hooks/useAgentSession.ts", import.meta.url), "utf8");
+  const start = session.slice(session.indexOf('case "agent_start":'), session.indexOf('case "agent_end":'));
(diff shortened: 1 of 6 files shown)


uvforce:
Addressed in 37b6553. When a turn starts, the client reads the runtime's thinking level from the agent state and adopts it, so the label reports the level actually in effect even after "Use pi default" left an earlier explicit level in place. The dropdown reflects the same value once the turn ends.

agegr (approved):
Small, focused, and answers a real question during a long turn. Approving.

I merged current `main` into the branch and tightened the turn-start fetch you added in 37b6553: it now captures `sid` and `promptRunIdRef` before the request and drops the response if either changed while it was in flight — the same guard `reconcileAgentState` uses — so a slow reply cannot set the level for a different session or a later run. Test updated to match.

Verified: tsc, eslint, `ChatInput*` and `hooks/*` tests pass.

PR #778, opened 2026-09-08
Make the session list and file explorer split resizable

uvforce:
## Problem

The sidebar splits its height evenly between the session list and the file explorer whenever the explorer is open. The split cannot be changed, so the balance is wrong for most of the work in the sidebar: browsing a deep file tree leaves half the sidebar showing sessions, and scanning a long session list leaves half of it showing a handful of files.

The sidebar width and the file panel width are both resizable, so this is the one panel boundary that cannot be adjusted.

## Change

The boundary between the session list and the explorer becomes a draggable separator, using the same `useResizablePanel` hook that already drives the sidebar and file panel. That means the same drag, `Home`/`End`, `Enter` to reset, shift-step, pointer capture, blur cancellation, and width reclamping on window resize, plus persistence under `pi-explorer-height`.

To reuse the hook, `growthDirection` accepts `up` and `down` in addition to `left` and `right`. The vertical axis reads `clientY`, uses `row-resize`, reports `aria-orientation="horizontal"` for the separator, and drives the arrow keys of that axis. Horizontal callers are unchanged.

The explorer keeps its own height and the session list takes the remaining space, with both sides floored so neither collapses. The height is applied through a CSS variable on the sidebar rather than the explorer element, because the explorer only mounts once a project is open.

## Tests

- `components/SessionSidebar.explorer-split.test.mjs` — the sidebar drives the shared resizer, and the hook handles both axes without changing horizontal behaviour.
- Verified in a production build with a real browser: dragging the separator moves the boundary (explorer 260 → 340, list 590 → 510), the value persists and is restored when the project is reopened, `Enter` resets to the default, arrow keys step it, and the existing sidebar width resizer still works (260 → 330).

### DIFF
--- a/components/SessionSidebar.explorer-split.test.mjs
+++ b/components/SessionSidebar.explorer-split.test.mjs
@@ -8,6 +8,10 @@ const hook = await readFile(new URL("../hooks/useResizablePanel.ts", import.meta
 test("the explorer split reuses the shared panel resizer", () => {
   assert.match(source, /useResizablePanel\(\{[\s\S]*?cssVariable: "--explorer-height"[\s\S]*?growthDirection: "up"[\s\S]*?storageKey: "pi-explorer-height"/);
   assert.match(source, /\{\.\.\.explorerResizer\.separatorProps\}/);
+  // The ceiling is the space the list and explorer share, not the whole sidebar:
+  // the header above the list is fixed, so counting it would let the explorer overflow.
+  assert.match(source, /const listHeight = listScrollRef\.current\?\.getBoundingClientRect\(\)\.height \?\? 0;/);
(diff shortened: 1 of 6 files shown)


uvforce:
Addressed in 31e5319. The ceiling now measures the session list and the explorer themselves rather than the whole sidebar, so the fixed header above the list is no longer counted as resizable space. Verified in a browser: with a 129px header the explorer stops at 826px (was 955px) and the bottom of the file tree is no longer clipped.

uvforce:
Closing this: I no longer run the feature locally, so I cannot keep validating it against upstream changes, and I would rather not leave a PR open that I am not exercising.

The branch still applies cleanly to v0.9.1 and CI was green, so if anyone wants the split, it is all here and free to pick up.

PR #785, opened 2026-09-09
Show the session message count in the header

uvforce:
## Problem

The header reports tokens, cost, and context usage, but not how many messages the session holds. That number is the one that predicts a specific problem: the session file grows with it, and past a few thousand messages opening or switching to that session is visibly slower. Today the only way to see it is the sidebar row, which is not visible while reading the conversation.

## Change

`renderSessionStatsButton` gains a message count next to the existing token and context readouts, using the same compact formatting (`1.2k`, `3.4M`) and the tooltip already assembled there.

It reuses the thresholds and colours of the context gauge: muted by default, amber past 2,000 messages, red past 5,000. Nothing is shown for an empty session, so a new chat keeps exactly the header it has today. The mobile toolbar is unchanged.

## Tests

- `components/AppShell.session-stats.test.mjs` — the count renders only for a non-empty session, joins the tooltip, and shares the warning thresholds with the context gauge.
- Verified in a production build with a real browser: after one exchange the header shows the count beside the token and context readouts, and the tooltip reads `messages: 2  |  in: 7,511  |  ...`.

### DIFF
--- a/components/AppShell.session-stats.test.mjs
+++ b/components/AppShell.session-stats.test.mjs
@@ -1,0 +1,22 @@
+import assert from "node:assert/strict";
+import { readFile } from "node:fs/promises";
+import test from "node:test";
+
+const source = await readFile(new URL("./AppShell.tsx", import.meta.url), "utf8");
+const stats = source.slice(source.indexOf("const renderSessionStatsButton"));
+
+test("reports the session message count alongside tokens and context", () => {
+  assert.match(stats, /const totalMessages = sessionStats\?\.totalMessages \?\? 0;/);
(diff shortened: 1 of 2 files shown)


agegr:
Thanks, but I'm going to pass on this one. The header stats are meant to stay at the few numbers that drive a decision (tokens/cost/context → compact or not). A message count with amber/red thresholds is not something the user can act on: if sessions with thousands of messages get slow to open, that is a pi-web performance problem to fix (paging already exists via `historyCursor`), not a warning to surface in the header. The sidebar row already shows the count for anyone curious.

Closing to keep the header lean.

Issue #671, opened 2026-08-31
Performance: allow a dedicated model and bounded context for session title generation


uvforce:
## Problem

Pi Web's session title generation can be unexpectedly slow and expensive for sessions using a large reasoning model or a large active context.

In `POST /api/sessions/[id]/auto-name`, the temporary title-generation agent currently inherits the active session's:

- model
- thinking level
- system prompt
- active conversation messages
- tool definitions

It also waits for the active agent to become idle and has a 90-second timeout.

As a result, generating an 8–24 character title may invoke the same large model and reasoning level used for the main coding task, potentially with tens of thousands of context tokens. This is disproportionate to the complexity of title generation.

## Proposed configuration

Please consider adding dedicated session-title generation settings, for example:

```json
{
  "sessionTitle": {
    "model": "current",
    "thinkingLevel": "off",
    "contextStrategy": "bounded",
    "timeoutMs": 15000
  }
}
```

Possible options:

### Model

- `current` for backward compatibility
- Any configured and authenticated model from the existing model registry

The Pi Web settings UI could expose this as:

> Title generation model: Current model / [available models]

### Thinking level

Allow `off` or `minimal` independently of the active session's thinking level.

### Context strategy

A bounded context should be sufficient for a short title, for example:

- the first user message
- the latest compaction summary, if available
- the most recent few user/assistant turns

This would avoid sending the full active coding context solely to generate a title.

### Timeout and fallback

If title generation times out or the configured model is unavailable, Pi Web could:

1. fall back to the current model, if explicitly enabled; or
2. derive a title from the first user message.

## Backward compatibility

The default model could remain `current`, preserving existing behavior unless the user selects a dedicated model.

Alternatively, a lightweight available model with thinking disabled could become the default for new installations.

## Suggested MVP

To keep the first implementation focused, the initial version could expose only:

1. Title generation model
2. Title generation thinking level

Bounded context, timeout configuration, and fallback behavior could follow separately.

## Benefits

- Faster title generation
- Lower token usage
- Avoids applying high reasoning levels to a trivial task
- Keeps the built-in Generate title button useful for long-running sessions
- Does not require users to patch Pi Web or maintain a separate extension

If this direction is acceptable, I would be happy to prepare a focused PR.

uvforce:
Implemented in #807.

[...]

PR #787, opened 2026-09-10
Generate session titles from a lightweight text transcript

uvforce:
## Problem

Generating a short session title currently replays the source system prompt, tool schemas, and conversation history, including tool results and reasoning. Tool-heavy sessions can therefore send substantially more context than is useful for naming.

## Change

The temporary title agent receives a short naming system prompt, no tools, and one user message containing a plain-text transcript followed by the existing title instructions. This transcript is extracted locally; there is no extra summarization-model call.

The transcript keeps messages in order and includes:

- Up to 800 Unicode code points from each nonempty user message.
- Up to 300 code points from each nonempty assistant text message, or 600 for the last nonempty assistant text message.
- Up to 600 code points from each compaction summary.

Tool calls, tool results, thinking blocks, images, and other message roles are omitted. All user turns with text remain represented, including changes of goal in the middle of a session. The obsolete shadow-tool, tool-pair sanitization, and trailing-user folding helpers are removed.

## Scope and relationship to #776

Related to #671. This PR is based directly on `main` and does not depend on #776 (dedicated title model selection). It changes the input, not the choice of model or thinking level. Both changes are complementary, but they touch the same implementation file and may need reconciliation when one lands.

The manual trigger, `waitForIdle()`, inherited model and thinking level, transport/credential hooks, 90-second generation timeout, title cleanup, usage reporting, and persistence route are unchanged. This PR does not add first-message automatic naming or any settings/UI changes.

## Trade-offs

The caps are heuristics, not a fixed total context or token budget: input still grows with the number of messages. Truncation can cut mid-sentence, and information present only in omitted tool/image content or after a cap is unavailable to the title model. The shorter prompt also stops deliberately matching the source conversation's cache prefix; this is not a claim of lower billed cost or latency for every provider/cache state. Custom conversion/context/payload hooks remain inherited as before.

## Validation

- The focused tests failed against the original implementation and now pass (9 tests).
- Provider-boundary tests use the real temporary Agent and SDK conversion with a synthetic streaming provider. They verify a single user request, absence of source-system/tool/image/thinking sentinels, source-message immutability, preserved model/transport choices, compaction-only history, unchanged idle waiting, Unicode caps and mid-session goals, usage reporting, and provider errors.
- Full `npm test`: 973 passed.
- TypeScript, ESLint, `git diff --check`, and production build passed.
- No live-provider quality/latency benchmark or browser end-to-end test was run for this isolated PR branch. No production session or settings were modified.

### DIFF
--- a/lib/session-title.test.mjs
+++ b/lib/session-title.test.mjs
@@ -6,11 +6,10 @@ import { createJiti } from "jiti";
 
 const jiti = createJiti(import.meta.url);
 const {
-  appendTitleRequestToTrailingUser,
   buildSessionTitleAgentOptions,
+  buildTitleTranscript,
   generateSessionTitle,
   parseGeneratedSessionTitle,
-  sanitizeTitleMessages,
 } = await jiti.import("./session-title.ts");
 
 function assistantMessage(text) {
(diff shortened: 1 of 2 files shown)


uvforce:
Superseded by #807, which adds a fixed total transcript budget, head/tail coverage, measured title-quality checks, and the dedicated title-model setting. Keeping the complete implementation in one PR avoids asking maintainers to reconcile two overlapping changes to the same title-generation path.

PR #790, opened 2026-09-10
Add a full-width toggle to the file panel

uvforce:
## Problem

The file panel is useful beside the chat, but its split width is restrictive when reading wide code, tables, documents, or interactive HTML previews. Dragging the separator helps only up to the layout's maximum width, and hiding the rest of the app manually would lose the convenient split layout.

## Change

Add an expand/restore button next to the existing hide-panel button.

- Expanded mode temporarily covers the app's available viewport without using the browser Fullscreen API.
- The existing file panel and `FileViewer` stay mounted, so interactive HTML state, file-viewer state, and scroll position are not reset by the layout switch.
- Restoring returns to the previous split width; the saved sidebar and file-panel widths are never overwritten.
- Hiding an expanded panel clears expanded mode, so reopening returns to the normal split.
- Background sidebar/chat controls and resize handles are `inert` while covered.
- The button is omitted on mobile, where the file panel already fills the viewport.
- Labels are included for English, Simplified Chinese, and Traditional Chinese.

The implementation is frontend-only and applies to the whole file panel rather than any one file type.

## Tests

- Added browser coverage at desktop and mobile widths using an interactive HTML fixture.
- Verified manual split resizing, repeated expand/restore cycles, keyboard activation, exact width restoration, unchanged localStorage widths, iframe identity, form state, scroll position, and hide/reopen behavior.
- `npm test`: 973 passed.
- TypeScript, ESLint, `git diff --check`, and production build passed.
- Full production-build E2E passed locally with system Chrome at 1280px and 390px.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -1676,6 +1676,48 @@ span.token.table {
   }
 }
 
+/* Full-width mode keeps the mounted viewer and saved split width intact. */
+@media (min-width: 641px) {
+  .right-panel-container.right-panel-open.right-panel-full-width {
+    position: fixed;
+    inset: 0 env(safe-area-inset-right) 0 env(safe-area-inset-left);
+    z-index: 250;
+    width: auto;
+    min-width: 0;
+    height: var(--app-viewport-height, 100dvh);
+    border-left: none !important;
+    box-shadow: none;
+    transform: none;
+    transition: none;
+  }
+  .right-panel-container.right-panel-full-width > * {
+    width: 100%;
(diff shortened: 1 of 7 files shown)


agegr (approved):
Compact and practical: one state flag, a CSS class that overlays the already-mounted panel, and `inert` on the covered regions so focus cannot land behind it. Saved widths untouched, mobile excluded. Merges cleanly onto current `main`; tsc, eslint, and the AppShell tests pass. Approving.

PR #792, opened 2026-09-10
fix: preserve the streaming partial when switching back to a running session

infinitex233:
## Problem

Switching away from a session whose model is still streaming and then switching back left the chat completely blank - no thinking text, no tool progress, and not even the phase label - until the current message finished.

## Root cause

Remounting ChatWindow for a still-running session raced two restore paths:

1. The SSE event stream reconnects and restores the in-flight partial from its snapshot (connected -> message_start). This is fast because the session is already alive server-side.
2. The mount effect's loadSession() resolves later (two sequential HTTP round trips: session file, then agent state) and unconditionally dispatched { type: "start" }, which resets streamingMessage to null and wipes the just-restored snapshot.

Every subsequent delta then no-oped in updateContentBlock (no baseline message) while each one also cleared the agent phase, so the UI showed nothing at all until the next message boundary.

## Fix

The mount restore now dispatches a reducer-owned "resume" action that enters streaming state without discarding a partial the event stream already restored. "start" keeps its reset semantics for genuinely new runs (send, agent_start). The partial is still owned by the single streamReducer - no state-mirroring refs.

## Tests

- lib/streaming-message.test.mjs: resume keeps a partial restored by a reconnect snapshot (identity preserved, deltas keep accumulating); resume enters streaming state when nothing was restored yet.
- hooks/useAgentSession.test.mjs: the mount restore dispatches resume, never start.

npm test (976 passing), npm run lint, and npx tsc --noEmit are green.

## Reproduction

1. Send a prompt that makes the model think/stream for a while.
2. Switch to another session, then switch back.
3. Before: blank until the next message boundary. After: the streaming bubble is restored immediately and keeps growing.

### DIFF
--- a/hooks/useAgentSession.test.mjs
+++ b/hooks/useAgentSession.test.mjs
@@ -533,3 +533,16 @@ test("keeps a detached viewport in place when streaming completes", () => {
   assert.doesNotMatch(scrollEffectSource, /\|\|/);
   assert.match(source, /addEventListener\("scroll", handleScrollPositionChange/);
 });
+
+test("mount restore resumes the stream instead of resetting a restored snapshot", () => {
+  const mountRestoreSource = source.slice(
+    source.indexOf("if (agentState.state?.isStreaming || agentState.state?.isPromptRunning) {"),
+    source.indexOf("if (agentState.state?.isBashRunning) {"),
+  );
+
+  // The mount restore runs after the event stream may already have restored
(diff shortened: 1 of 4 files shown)


agegr:
Closing as superseded. The same streaming-resume fix is already included on main in commit 4787a14 ("fix(ui): preserve streaming output when opening active session"), so this PR no longer needs to be merged.

Issue #795, opened 2026-09-10
Reading a session that another pi process is writing: stale refreshes, and follow-the-focused-session semantics


Nuctori:
# Session refresh cannot see external updates — and "follow the focused session" as the fix

## Summary

Following up on #632 (closed as NOT_PLANNED). I'm not asking to revisit cross-process write
coordination — I agree that needs conflict handling and isn't worth it.

But #632 bundled two separate things, and only one of them was the expensive one:

1. **A concrete, reproducible bug**: the refresh button cannot see external updates at all.
   This has nothing to do with cross-process coordination — it's a read-path bug.
2. **An undefined product decision**: what should "this session is focused in pi-web while
   another pi process is also using it" *mean*? Today it's undefined, so the behaviour is
   accidental rather than chosen.

The reporter's key observation ("the refresh icon doesn't trigger an update either") points at
(1), which is fixable on its own.

---

### Root cause

`GET /api/sessions/[id]` short-circuits to the in-memory session when a live wrapper exists
(`app/api/sessions/[id]/route.ts`):

```ts
const rpc = getRpcSession(id);
const liveRpc = rpc?.isAlive() ? rpc : undefined;
// ...
const sm = liveRpc?.inner.sessionManager ?? SessionManager.open(resolvedPath!);
const entries = sm.getEntries();   // <- from memory, never from disk
```

Pi's `SessionManager` builds its lookup index once, when the file is loaded
(`dist/core/session-manager.js:673 _buildIndex`), and appends with `appendFileSync` (`:753`).
So:

- **pi TUI appends** → file changes on disk
- **pi-web's wrapper** still serves the entries it indexed at load time
- **the refresh button re-requests the same shadowed endpoint** → it cannot help, by construction

The stale snapshot persists until the wrapper is evicted by the idle timeout
(`DEFAULT_SESSION_IDLE_TIMEOUT_MS = 10 * 60 * 1000`), which is why a manual reload
"eventually" works.

### Why this isn't the declined scope

This is a **read-path** bug. No writes from pi-web, no coordination, no conflict resolution.
Making the read observe what is already on disk is strictly a correctness fix for the read.

### Reproducible

1. Open session X in pi-web (this creates a live wrapper)
2. Append one entry to X's `.jsonl` from another process (what the pi TUI does)
3. `GET /api/sessions/X` → the new entry is **absent**
4. Still absent after `?force=1`, after reloading the page, and after the refresh button
5. It appears only after the 10-minute idle eviction

[...]

Issue #632, opened 2026-08-27
pi-web UI does not auto-refresh when Pi session output changes


kafumanto:
Hi! I’m a new user and I’m not sure whether this is the intended behavior or a bug.

I’m running Pi and pi-web inside a VS Code Dev Container on WSL. VS Code automatically forwards the pi-web port, and when I open the forwarded port from my Windows host I can see active Pi sessions and interact with them normally.

However, while Pi continues updating its output, pi-web does not automatically refresh the displayed content. The page only updates if I manually reload it in the browser (F5 in Chrome). The refresh icon inside pi-web doesn’t trigger an update either.

Is this expected, or should pi-web auto-refresh the session output?

Thanks for the great tool!

kafumanto:
I just noticed that also the opposite is true apparently: a change from pi-web is not reflected in Pi automatically, I must close and reopen Pi and then use `/resume` to see the changes from pi-web.

Likely it's just a common limitation, but it could be a nice feature request :) Feel free to close this if it's not a bug and the feature is not planned. Thanks!

Nuctori:
+1， same to me

aievy:
this also happens for me. is this supposed to be like that? i would expect changes in a session to appear instantly on both the console and web view of the session. i don't use any proxy / forwarding, just plain powershell and local pi-web instance.

agegr:
Pi Web supports browsing and resuming Pi session files, but concurrent use of the same session by Pi and Pi Web is not supported. Each process maintains its own in-memory session state, so live bidirectional synchronization would require cross-process coordination and conflict handling. We do not plan to add that complexity.

PR #796, opened 2026-09-10
fix: see sessions written by another pi process

Nuctori:
## What this fixes

`GET /api/sessions/[id]` short-circuits to the in-memory session whenever a live wrapper
exists, so the entries it returns come from memory and never from disk:

```ts
const rpc = getRpcSession(id);
const liveRpc = rpc?.isAlive() ? rpc : undefined;
// ...
const sm = liveRpc?.inner.sessionManager ?? SessionManager.open(resolvedPath!);
const entries = sm.getEntries();   // from memory, never from disk
```

Pi's `SessionManager` builds its lookup index once at load time
(`dist/core/session-manager.js:673 _buildIndex`) and appends with `appendFileSync` (`:753`).
So when the pi TUI writes the same session file, the wrapper keeps serving the snapshot it
indexed at load, and the new output stays invisible until the wrapper is evicted by the idle
timeout (`DEFAULT_SESSION_IDLE_TIMEOUT_MS`, 10 minutes).

The refresh button cannot help **by construction** — it re-requests the same shadowed endpoint.
That also makes `?force=1` ineffective, which is what made this look like "no auto-refresh" in
#632 when reloading did eventually recover.

### Reproduction

1. Open session X in pi-web (creates a live wrapper)
2. Append one entry to X's `.jsonl` from another process (what the TUI does)
3. `GET /api/sessions/X` → the new entry is **absent**
4. Still absent after `?force=1`, a page reload, and the refresh button
5. It appears only after the 10-minute idle eviction

## The change

Detect that the newest on-disk entry id is unknown to the wrapper, then drop the wrapper so the
request rebuilds from the file:

```ts
let liveWrapper = rpc?.isAlive() ? rpc : undefined;
if (liveWrapper && !liveWrapper.isRunning()) {
  const diskLatestId = readLatestSessionEntryId(liveWrapper.sessionFile);
  if (diskLatestId && !liveWrapper.inner.sessionManager.getEntry(diskLatestId)) {
    liveWrapper.destroy();
    invalidateSessionListCache();
    liveWrapper = undefined;
  }
}
```

Three details worth noting:

- **Idle only.** Mid-run the wrapper owns the write path, and an external writer during an
  active run is the genuinely unsupported concurrent-write case rather than a stale read. A
  session blocked on user input also has an unsettled prompt, so it is not evicted.
- **Rebuilding is safe.** Eviction lands on the ordinary cold-start path:
  `startRpcSession` restores the persisted tool selection from the file
  (`readSessionToolSelection`) and the model from its `model_change` entries — the same path
  already taken after idle eviction or a restart.
- **The probe is cheap and defensive.** It reads a bounded 64 KB tail, skips the session
  header (which carries the session *id*, not an entry id, and is excluded from the SDK's entry
  index — treating it as an entry would evict a fresh wrapper on every read), and tolerates a
  torn trailing line, since `appendFileSync` is not atomic from a reader's perspective.

## Why this isn't the scope declined in #632

#632 was closed because live **bidirectional** synchronization would need cross-process
coordination and conflict handling. I'm not proposing that, and this PR adds none of it:

- No writes from pi-web, no locking, no ownership protocol, no conflict resolution.
- It only makes an existing read observe what is **already on disk**. Nothing is synced; the
  wrapper simply stops lying about the file.

**Explicitly not in scope:**

- Auto-push/`tail -f` for a focused session (a follow mode with an SSE watcher). That is a new
  capability, not a correctness fix, and it deserves its own discussion.
- The pi-web → pi direction (the reporter's "I must close Pi and `/resume`"). pi never watches
  session files, so that is an upstream `pi-mono` gap rather than a pi-web bug.

## Verification

```
npm run lint         0 errors (1 pre-existing warning in e2e_click.mjs)
npx tsc --noEmit     clean
npm test             43 pass / 0 fail across the touched suites
node e2e/run.mjs     PASS: external session-file appends are visible despite a live wrapper
```

Regression coverage added:

| File | Covers |
| --- | --- |
| `lib/session-reader-latest-entry.test.mjs` | newest id, header skip, torn trailing line, absent file, bounded read on a 2 MB file, CRLF |
| `lib/session-reader-external-write.test.mjs` | the decision on a real `SessionManager`: an externally appended entry is unknown, while the manager's own append is always known (no self-eviction loop) |
| `e2e/run.mjs` | warms a real wrapper, appends as the TUI would, asserts the entry becomes visible and that repeated reads stay stable |

**Negative control** (fix stashed, test kept) fails with the exact symptom of the report:

```
AssertionError: an external append must become visible
  [ 'root', 'reply' ]        # 'external' missing
```

## Notes

- I could not run `npm run build` locally: this Windows machine fails the webpack build with
  `EPERM: readlink 'C:\Users\Nuctori\AppData\Local\Intel\SUR\QUEENCREEK\...'` (an unrelated
  Intel driver directory), and an aggressive scanner keeps emptying `node_modules` (selectively,
  native/crypto packages) during installs. Both are machine-local; CI runs on `ubuntu-latest`.
  The `E2E_SERVER_MODE=dev` run above passes, including the new regression test.
- One pre-existing failure is visible in the local dev-mode E2E:
  `/api/sessions/e2e-compacted-session/state` returns 500 at the 390px viewport. It is
  unrelated to this change — it reproduces identically with this branch stashed, and main's CI
  is green, so it appears to be Windows-specific.

Related: #632 (closed).

Follow-up discussion on follow-the-focused-session semantics: #795.

### DIFF
--- a/app/api/sessions/[id]/route.ts
+++ b/app/api/sessions/[id]/route.ts
@@ -10,6 +10,7 @@ import {
   invalidateSessionListCache,
   buildSessionContext,
   readSessionHeader,
+  readLatestSessionEntryId,
 } from "@/lib/session-reader";
 import { sessionPathKey } from "@/lib/session-path";
 import { getRpcSession } from "@/lib/rpc-manager";
@@ -27,7 +28,25 @@ export async function GET(
   const { id } = await params;
   try {
     const rpc = getRpcSession(id);
-    const liveRpc = rpc?.isAlive() ? rpc : undefined;
+
+    // A live wrapper only reflects the appends pi-web itself made. When another
+    // pi process (the TUI) writes the same session file, the wrapper's in-memory
+    // entries silently diverge from disk and every read — including a manual
+    // refresh — keeps serving the stale snapshot. Detect the newest on-disk
+    // entry the wrapper never saw and drop the wrapper, so this request rebuilds
+    // from the file instead of from memory. Only while idle: mid-run the wrapper
+    // owns the write path, so an external write then is the genuinely
+    // unsupported concurrent-write case rather than a stale read.
+    let liveWrapper = rpc?.isAlive() ? rpc : undefined;
+    if (liveWrapper && !liveWrapper.isRunning()) {
+      const diskLatestId = readLatestSessionEntryId(liveWrapper.sessionFile);
+      if (diskLatestId && !liveWrapper.inner.sessionManager.getEntry(diskLatestId)) {
+        liveWrapper.destroy();
+        invalidateSessionListCache();
+        liveWrapper = undefined;
+      }
+    }
+    const liveRpc = liveWrapper;
     const resolvedPath = liveRpc ? null : await resolveSessionPath(id);
     if (!liveRpc && !resolvedPath) {
       return NextResponse.json({ error: "Session not found" }, { status: 404 });
--- a/e2e/run.mjs
+++ b/e2e/run.mjs
@@ -2,7 +2,7 @@
 import assert from "node:assert/strict";
 import { spawn } from "node:child_process";
 import { once } from "node:events";
-import { createWriteStream, existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { appendFileSync, createWriteStream, existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { createServer } from "node:net";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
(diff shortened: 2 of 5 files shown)


agegr (approved):
Confirmed the bug: `GET /api/sessions/[id]` serves `liveRpc.inner.sessionManager.getEntries()` whenever a wrapper is alive, and the SDK's `SessionManager` never re-reads its file, so appends from the TUI stay invisible until the 10-minute idle eviction — refresh and `?force=1` included.

The fix is well scoped: a bounded 64 KB tail probe, header/torn-line aware, and the wrapper is only dropped while idle (`!isRunning()`), which is the same lifecycle path fork already uses. Tests cover the header-vs-entry distinction and the torn trailing line. Merges cleanly onto current `main`; tsc, eslint, and `lib/session-reader*` tests pass.

Approving. Thanks for the precise diagnosis.

PR #801, opened 2026-09-11
feat: edit workspace Markdown files

QAyong:
## Summary\n\n- add Markdown editing for .md and .mdx files in the workspace file viewer\n- add a guarded PATCH file API with workspace-root, regular-file, JSON, and size checks\n- support Save, Cancel, Ctrl/Cmd+S, live-preview refresh, and localized edit labels\n\n## Verification\n\n- node_modules/.bin/tsc --noEmit\n- npm run lint\n- targeted Markdown route and FileViewer tests\n\nThe pre-existing unrelated working-tree changes in app/globals.css and components/SessionSidebar.tsx are not included.


agegr:
Thanks for the contribution. Closing this one on scope:

- The file panel in pi-web is a read-only viewer on purpose. Editing files is the agent's job (through pi's `edit`/`write` tools, which the chat then renders as diffs); adding a `PATCH /api/files` write path and an in-browser editor turns the viewer into a second editing surface that would need its own conflict handling, permissions, and UI, and moves pi-web away from being a thin wrapper around pi.
- The branch also carries the Windows launcher / provider-icon changes from #799, which I declined separately.

I'll close #838 (the larger follow-up of the same feature) for the same reason.

Issue #803, opened 2026-09-11
perf(sessions): session list and context have no ETag, so unchanged data is re-transferred on every load


YsDirector:
## Problem

`/api/sessions` and `/api/sessions/[id]` never send `ETag`/`Last-Modified`, and the list route also sends `Cache-Control: no-store`. A client with an unchanged session store therefore re-transfers the full payload on every page load, session switch, and app resume.

This is most visible on mobile. An installed PWA or a WebView shell does not keep the SPA alive across launches, so the app re-mounts on every open and re-fetches the session list plus the active session's context window each time. Measured on a local server with 82 stored sessions:

| request | body | after #731 (gzip) |
|---|---|---|
| `GET /api/sessions` | 95,446 B | 27,625 B |
| `GET /api/sessions/{id}?deferThinking=1&deferMedia=1` (8,712-message session, `tail=50` page) | 257,987 B | 100,591 B |

gzip (#731) removed ~71%, but nothing lets the client skip a request whose answer has not changed:

```
$ curl -D - -H 'If-None-Match: "anything"' .../api/sessions
HTTP/1.1 200 OK
cache-control: no-store
content-encoding: gzip
```

## Why this should be cheap to fix

The server already computes exactly what a validator needs:

- `app/api/sessions/route.ts` returns `sessionListVersion` (`getSessionListVersion()` → `globalThis.__piSessionListGeneration`), a monotonic counter the client already uses in `components/SessionSidebar.tsx` to decide whether to refresh.
- `app/api/sessions/[id]/route.ts` already calls `statSync(filePath).mtime` to populate the `modified` field.

So a weak `ETag` is derivable on both routes without adding I/O.

## A wrinkle: `no-store` blocks revalidation

The routes forbid storing the response, and the client requests the list with `fetch(..., { cache: "no-store" })` in two places. A browser that may not store a response cannot send `If-None-Match`, so adding an `ETag` alone would change nothing. Making 304s reachable means relaxing these to `no-cache` (store, but always revalidate). That is a real API behavior change, so I am opening an issue rather than sending a PR.

## Related: two call sites pull the whole list for one session

- `components/AppShell.tsx:570` (`restoreWorkspaceContext`) — finds the last-open session id.
- `components/AppShell.tsx:768` (`hydrateSelectedSession`) — hydrates one transient session.

Conditional requests would make both free when nothing changed; a targeted lookup (e.g. `?id=`) would avoid the transfer entirely.

## Proposal

[...]

PR #805, opened 2026-09-11
feat(pwa): serve the cached app shell when the network is unreachable

YsDirector:
## Summary

Reopening the app without a connection currently shows the `offline.html` placeholder, even though the service worker has already cached every static asset. Navigations are pure network-first with a single fallback to that placeholder, and the app shell document itself is never stored.

This caches the shell document and replays it offline, so the app boots from cache.

- cache the app shell document per navigation, then serve it when the network fails
- prime the shell cache during `install`
- bound navigation/asset fetches with an 8s timeout
- never store a redirect body under the requested URL
- `/api/` handling and the push handlers are untouched

## Why prime the shell cache in `install`

The navigation that registers the worker cannot be captured — the worker is not controlling the page yet, so its `fetch` handler never sees that response. Without priming, a user who opens the app once and then goes offline still lands on `offline.html` on the next launch. This is easy to miss when testing by loading the page twice.

Priming is best-effort: an unreachable server, a redirect to the login page, or a non-HTML response leaves the shell cache empty rather than failing the install.

## Why the timeout

A reachable port with a dead upstream (a stopped dev server behind a proxy, a tunnel that is up while the remote is down) never rejects. `fetch().catch()` therefore never runs, the navigation hangs, and the user gets a blank page instead of the offline fallback.

## Behavior

| URL | Online | Offline |
|---|---|---|
| `/` | network (cached as shell) | cached shell |
| `/?session=x` | network | cached shell |
| `/_next/static/*` | network, falls back to cache | cache |
| `/api/*` | network (unchanged) | network error (unchanged) |

Offline, the router reads the address bar itself, so serving the cached shell for any in-app URL still resolves the right session once data is available.

## Testing

```
node --experimental-strip-types --test public/sw.test.mjs   # 11 passed
env PI_CODING_AGENT_DIR=/tmp/pi-web-test-agent npm test     # 1026 passed
node_modules/.bin/tsc --noEmit
npm run lint
```

Six new tests cover offline shell replay, shell reuse across query strings, install priming, install against an unreachable server, redirect safety, and that `/api/` traffic stays untouched. The existing push and notification-click tests are unchanged.

## Scope

Deliberately limited to the shell. Caching session JSON for offline history is a larger behavioral change (it alters what `/api/` may serve from cache) and I would rather raise it separately if this direction looks right.

Related: #803 (conditional requests for `/api/sessions`) is independent — this PR is about working offline at all, that one is about not re-transferring unchanged data while online.

### DIFF
--- a/public/sw.js
+++ b/public/sw.js
@@ -1,6 +1,8 @@
 const CACHE_PREFIX = "pi-web";
 const CACHE_VERSION = new URL(self.location.href).searchParams.get("v") || "dev";
 const STATIC_CACHE = `${CACHE_PREFIX}-static-${CACHE_VERSION}`;
+const SHELL_CACHE = `${CACHE_PREFIX}-shell-${CACHE_VERSION}`;
+const CURRENT_CACHES = new Set([STATIC_CACHE, SHELL_CACHE]);
 const OFFLINE_URL = "/offline.html";
 const PRECACHE_URLS = [
   OFFLINE_URL,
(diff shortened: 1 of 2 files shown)


agegr:
Thanks for the thorough write-up — the reasoning about priming in `install` and the hung-upstream case is correct.

I'm still going to close this on product grounds. Every screen in pi-web needs the server: the session list, the chat history, the file panel, and sending a prompt are all `/api/*` calls. Booting a cached app shell offline therefore does not give the user a usable app; it gives them a shell whose every request fails, which is a worse signal than the honest `offline.html` placeholder we show today. It also adds the classic PWA failure mode (a cached HTML document referencing `_next/static` chunks that are no longer on the server after an update) and roughly doubles the size of the service worker to manage.

The one part I think is worth landing on its own is the **bounded navigation fetch**: a reachable port with a dead upstream currently hangs instead of falling back to `offline.html`. If you open a small PR that only adds the timeout around the existing navigation/asset fetches (no shell caching), I'll review it.

PR #807, opened 2026-09-12
Name sessions from a bounded transcript

uvforce:
Supersedes #787 and replaces the approach proposed in #776.

Closing #776 was the right call for what that PR contained. Switching to a dedicated model could lose prompt-cache hits because the title request still replayed the system prompt, tool definitions, and full message history. A cheaper model was therefore not necessarily a cheaper request.

This PR fixes that underlying cause: title generation no longer sends the session prefix. The request has a fixed total transcript budget, so its cost and latency stop depending on session length or cache state.

## What changes

`generateSessionTitle` builds a bounded plain-text transcript and sends it as a **standalone stream**, the same request shape compact uses for one-off summaries:

- No system prompt, tool definitions, tool calls, or tool results.
- User turns carry more budget than assistant turns; the final assistant turn keeps extra room for the outcome.
- The total budget is split between the opening and newest turns, with an elision marker between them. A 12-turn and a 1200-turn session produce the same bounded request.
- Thinking uses the cheapest level returned by `getSupportedThinkingLevels()`.
- The request uses `cacheRetention: "none"` and a fresh session id. It does not clone a throwaway Agent or reuse the live session's `transformContext` / `onPayload` hooks.
- Image-only turns keep an `[image]` placeholder. A session with no usable text fails before the model is called.

Default model selection is unchanged: naming uses the session's own model. There is no Title model setting.

## Measured results

Using the same session file and session model (`claude-haiku-4-5`), changing only the title-generation implementation:

- At 120 turns, input fell from about **74k tokens to 1.7k**.
- Latency remained around **1.3 seconds** instead of rising to 3.8 seconds.
- All **15/15 adversarial quality runs** still named the actual subject, including mid-session pivots and subjects buried under a long tail of unrelated chores.

## Behaviour and scope

- Naming no longer waits for the agent to become idle. Because there is no session prefix to append to, a running session can be named from its current transcript snapshot.
- This PR does **not** automatically name a new session after its first message; that behaviour is outside this PR.
- Addresses #671.


agegr (approved):
The bounded-transcript change is the part that matters here and it is a clear win: a title request that no longer replays the system prompt, tools and full history is cheaper on *any* model, and the head/tail split with per-role caps is a sensible way to keep pivots visible. Measured numbers are convincing. Thanks for persisting on this.

I pushed a commit to the branch that keeps exactly that and removes the **Title model** setting (settings route, `lib/session-title-settings.ts`, the General settings section, CSS, i18n, and the `options.model` / `resolveTitleModel` path in `lib/session-title.ts`). Same reasoning as on #776: once the request is ~1–2k tokens flat, a dedicated cheaper model saves very little, and pi-web tries to keep the Settings surface small. The session's own model, with the cheapest thinking level it supports, is the single default. Tests were adjusted accordingly (`resolveTitleThinkingLevel` coverage kept).

Net diff is now `lib/session-title.ts` + its test. Merged current `main`; tsc, eslint, and the session-title / SettingsPanel / AppShell.auto-name tests pass. Approving.

PR #809, opened 2026-09-12
fix: close SSE streams on shutdown so Next 16 drain cannot strand a zombie node

cy7372:
## Problem

In production, Next 16 handles SIGINT/SIGTERM by calling `server.close()` and then waits **indefinitely** for every connection to end — no timeout, and `closeAllConnections` is only wired in dev mode. An SSE stream never ends on its own (it waits for the *client* to disconnect), so on any service restart the process lingers forever in a half-dead state:

- the listener is closed → reverse proxy sees **502**
- but the node process never exits → the supervisor never sees a failure and never restarts it
- every restart leaks one more orphaned node

(Found during recurring "zombie node, 502" outages on a systemd-managed deployment; same failure mode applies to any supervisor that stops the app with SIGTERM.)

## Fix

- `instrumentation.ts`: register SIGINT/SIGTERM hooks that close all live agent event streams, so Next's own drain finishes and the process exits cleanly
- `lib/agent-event-stream.ts`: registry of live stream closers. The registry is shared via `Symbol.for` + `globalThis` because `instrumentation.ts` and the route handlers are bundled into **separate module graphs** — a plain module-level `Set` would be two disconnected copies (verified live: the shutdown hook closed an empty set while the route's SSE kept heart-beating)

No behavior change for normal operation; streams close as they would on client disconnect.


agegr (approved):
Verified against `next@16.3.5`: `start-server.js` cleanup awaits `server.close()` and only calls `closeAllConnections()` when `isDev`, so a production process with an open SSE stream indeed never finishes draining. The `Symbol.for` registry is the right call given the separate module graphs, and erroring the controller (rather than `close()`) is what actually tears the socket down.

I merged `main` and pushed a small cleanup: trimmed the date/"Servy"-specific comments to the durable explanation and removed the unused `activeAgentEventStreamCount` export. Behavior unchanged; tsc, eslint, and the agent-event-stream / agent route tests pass. Approving.

PR #810, opened 2026-09-12
fix: count only visible messages toward the session tail budget

cy7372:
## Problem

The session tail window was counted in **raw entries**, so toolResult and other non-rendering entries consumed the budget. In agent-heavy sessions the window starved user messages out of the chat: a real session with 15 user messages across 562 entries loaded exactly **one** user message; every earlier question needed manual "load earlier" paging.

## Fix

`countsTowardTail()`: only user / assistant messages plus the compaction divider consume the tail budget; toolResult entries, hidden custom messages and session meta ride along inside the span without consuming it. Older history still loads on demand via pagination.

Includes a regression test for the pagination path.


agegr (approved):
Confirmed: with `tail` counted in raw entries, a tool-heavy session (user → assistant+toolCall → toolResult → assistant, ×N) fits roughly one user turn into the default 50-entry page. Counting user/assistant messages plus the compaction divider is the right unit, and `hasMore` keeps paging working.

I merged `main` and pushed two adjustments:

- **Raw cap lowered** from `max(1500, tail·30)` to `max(200, tail·6)`. Tool results can be large, and 1500 raw entries per page (45 000 at `tail=1000`) undoes the payload bound this slice exists for. With a 1:1 assistant:toolResult ratio `tail·6` is still 3× headroom, and the search path (`tail: 200`) gets 1200. Added a test for the cap.
- Restored the `sliceActiveBranch` JSDoc (the new helper's comment had been inserted between the doc block and the function) and dropped the dated anecdote from the comment.

tsc, eslint, and `lib/session-reader*` + `app/api/sessions/**` tests pass. Approving.

PR #811, opened 2026-09-12
fix: use crypto randomUUID for manual-code handshake tokens

cy7372:
The OAuth manual-code flow returns a short-lived token used to authorize the pending login. Tokens were derived from `Math.random()`, so an attacker who knew the provider and the timing window could predict the token and complete someone else's flow. Use `crypto.randomUUID()` instead.


agegr (approved):
Correct and minimal: the manual-code token gates completion of a pending OAuth flow, so it must be unguessable. `randomUUID()` keeps the `<provider>-` prefix the POST handler checks. Approving.

PR #812, opened 2026-09-12
feat: awaiting-input state, compaction message queueing, richer extension dialogs

cy7372:
## Summary

Sessions blocked on an extension `ui_request` now read as **awaiting-input** instead of running, and prompts submitted during context compaction **queue** instead of erroring.

### Awaiting-input state

- `/api/agent/running` exposes `awaitingInputSessionIds`; the sidebar and project dropdown render an awaiting badge (amber) distinct from the running spinner
- the wrapper parses `overlayOptions.awaiting` on custom overlays: interactive dialogs count toward the awaiting badge and attention notifications, while passive overlays (toasts, footer stats) stay silent and render as collapsed pills
- interactive panels (flagged `awaiting`, or carrying touch actions) auto-expand; passive ones start collapsed
- browser notifications gated on `awaiting` for `method: "custom"`

`overlayOptions.awaiting: true` is a convention extensions can adopt to flag blocking dialogs — the TUI ignores unknown keys, so it is backward compatible.

### Compaction message queueing

The SDK rejects prompts during compaction ("Cannot submit a prompt while compaction is in progress"), which made the web input error out mid-compaction. Now:

- prompts submitted during compaction park server-side, surface as queued rows via merged `queue_update` events, and are recallable with `clear_queue`
- on `compaction_end` they flush FIFO; when auto-compaction ends back into a live run they join it as `followUp`
- parked prompts keep the wrapper marked running so idle cleanup cannot swallow them
- compacting status row above the input with queued count (en/zh-CN/zh-TW)

### Dialog rendering

- touch-action row mapping in the custom panel (`normalizeCustomPanelLinesMapped` maps action row numbers to display lines after frame/blank-line trimming; unmapped actions fall back to bottom buttons)
- compaction summary card is expandable with file lists and clickable file links

## Notes

- Also carries small test-hygiene fixes (CRLF-normalized source reads, SAFETY comments on legacy pi-ai flat image casts) so the suite passes on Windows checkouts.
- Verified: `tsc --noEmit` clean, 957 tests pass.


agegr:
Thanks — there is a real fix in here, but as a whole I can't take this PR, so I'm closing it and asking for one focused follow-up.

**Compaction message queueing** is the part I want. The SDK rejecting prompts during compaction and the web input erroring out is a genuine bug, and parking them server-side and flushing FIFO on `compaction_end` is the right fix. Please open that as its own PR (rpc-manager + `queue_update` merging + the status row), rebased on current `main` — this branch is conflicting.

The other two parts I'd decline:

- **Awaiting-input state / `overlayOptions.awaiting`** — this invents a pi-web-specific convention that extensions would have to adopt for the badge to be correct. pi-web should render what pi's extension protocol already expresses, not define new keys extensions need to learn about; an "awaiting" badge that is only right for extensions that know about pi-web is misleading for everyone else.
- **Richer dialog rendering** (touch-action row mapping, auto-expand heuristics) — `main` has had several rounds of dialog changes since this branch (fcd94bf, ed0eea9, #699, #612). Any further dialog work needs to start from that, and honestly I'd like the dialogs to settle rather than grow.

24 files / 1250 lines across three unrelated concerns is also just too much to review as one unit.

PR #813, opened 2026-09-12
Remember the selected thinking level for new sessions

Sure-Will:
Fresh composers currently reset the reasoning selector even after the user explicitly chose a level in the previous session. This stores the last explicit choice (including `auto`) in browser storage, reapplies it only to fresh sessions after model-scope compatibility checks, and keeps existing sessions bound to their own persisted state.

Validation:

- `npm test` (1031 passed)
- `npm run lint`
- Playwright at 1280px and 390px: refresh, old-session isolation, incompatible-model fallback, and `auto`/`off`


agegr:
Thanks, but this duplicates behavior `main` already has on the server side. When you pick an explicit thinking level for a new session, `lib/startup-preferences.ts` persists it as pi's default (`settingsManager.setDefaultThinkingLevel`), exactly like choosing a level in the pi TUI does. The next fresh composer shows `auto`, and `auto` *means* "pi's current default", which is the level you last chose — so the effective level is already remembered.

Adding a second, browser-local memory of the same choice (with its own compatibility checks and fallbacks) gives us two sources of truth that can disagree (different browsers, TUI changing the default in between), which is what the `auto` placeholder was designed to avoid. Closing.

PR #814, opened 2026-09-12
Show the selected thinking level beside assistant model names

Sure-Will:
Assistant replies currently show only the model, so historical turns lose the reasoning level selected when Send was pressed. This adds the display label beside the model, records a bounded versioned custom entry after prompt preflight accepts, and restores exact `auto` and provider-mapped labels across streaming reconnects and history reloads while legacy replies use branch setting history.

Validation:

- `npm test` (1034 passed)
- `npm run lint`
- Full Playwright E2E at 1280px and 390px


agegr:
Thanks for the thorough work. I'm going to close this one on cost/benefit.

Showing the thinking level next to the model on each reply is a small piece of metadata, and the implementation to make it exact touches 15 files: a new versioned custom entry written to the session file after every prompt preflight, changes in `rpc-manager`, `agent-event-stream`, `session-reader`, `useAgentSession`, and a reconnect/restore path. pi-web tries to stay a thin wrapper around pi and to avoid writing its own bookkeeping into pi's session files unless there is no alternative.

If this ever comes back, the acceptable shape is read-only: pi already records `thinking_level_change` entries in the session, so a label can be derived from branch history alone (your "legacy replies" path) with no new writes and no streaming plumbing. I don't think the label is worth more than that.

PR #815, opened 2026-09-12
Show total turn duration in existing chat metadata

Sure-Will:
Long turns provide no durable indication of how long the full run took. This measures from accepted prompt to logical idle across retries, tools, queued follow-ups, and compaction, persists a versioned record, and renders only the total duration in existing phase/process/message metadata rows. It intentionally adds no TTFT, token-rate metrics, or extra summary row.

Validation:

- `npm test` (1032 passed)
- `npm run lint`
- Playwright at 1280px and 390px: live ticking, completed/process turns, provider errors, empty Stop, pagination, and legacy sessions


agegr:
Thanks — same conclusion as on #814, for the same reason.

The chat already shows durations where they are actionable (thinking blocks, tool calls, the live phase indicator). A persisted total-turn duration adds ~800 lines across 20 files, including a new versioned record written into the session file, stream events, and reconnect handling, for one extra number in the metadata row. That is more machinery than pi-web wants to own for the value it adds, so I'm closing this rather than asking for changes.

PR #817, opened 2026-09-12
fix(plugins): invoke npm-cli.js on Windows for update checks

### LINKED ISSUE #742: [Windows] 0.9.0 Settings plugin "Check for updates" always fails with spawn npm ENOENT
Version: @agegr/pi-web 0.9.0 (core 0.85.1)
OS: Windows 11 IoT Enterprise LTSC (build 26100), Node v24.19.0, npm 12.0.2

Steps: Settings → Plugins → Check for updates
Result: always fails with `spawn npm ENOENT`, backend console is clean.
0.8.x worked, 0.9.0 broke.

Cause: `.next/server/app/api/plugins/check/route.js` calls bare
`execFile('npm', ...)` with no shell. On Windows npm is `npm.cmd`,
so it can never spawn. Reproduced locally:
`execFile('npm',['--version'])` → `spawn npm ENOENT`.
Upstream same-root-cause: earendil-works/pi#1220 #4665 #4793.

Expected: use `npm.cmd`/cross-spawn (like core's `spawnProcess`)
or honor `npmCommand` with shell on win32.

Workaround: set `"npmCommand": ["node", "<npm-cli.js>"]` in settings.

richardmilles:
execFile cannot spawn npm.cmd on Windows without a shell. Resolve npm-cli.js next to Node, matching lib/npx.ts.

Fixes #742

### DIFF
--- a/lib/plugin-updates.ts
+++ b/lib/plugin-updates.ts
@@ -1,6 +1,7 @@
 import { execFile } from "child_process";
 import { existsSync, readFileSync } from "fs";
-import { join } from "path";
+import { dirname, join } from "path";
+import { execPath } from "process";
 import { promisify } from "util";
 import {
   DefaultPackageManager,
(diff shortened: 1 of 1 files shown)


agegr:
Thanks for the fix — the diagnosis matches #742 exactly. I'm closing this in favor of #837, which fixes the same `spawn npm ENOENT` but does it by extracting the existing `findNpxCli` from `lib/npx.ts` into a shared `lib/node-cli.ts` used by both call sites, rather than adding a second copy of the finder in `lib/plugin-updates.ts`. Same behavior, one implementation, with a unit test.

PR #820, opened 2026-09-12
docs(AGENTS): sync API file map with app/api routes

richardmilles:
Add missing high-traffic routes (web-auth, plugins/check, project-trust, sessions/search, terminal, cwd/browse) to the File Map.


agegr (approved):
Thanks for syncing this. I pushed a follow-up to the branch: `web-auth` also has `GET` (status), `terminal/route.ts` is `POST` only with the per-terminal `GET/POST/DELETE` living in `terminal/[id]/route.ts`, and the remaining routes that were still missing (`sessions/[id]/state`, `sessions/[id]/auto-name`, `app-update`, `file-index`, `git/*`, `provider-usage/query`, `push/*`, `tools/settings`) are listed now so the map matches `app/api` fully. Approving.

PR #821, opened 2026-09-12
docs(AGENTS): list plugin-updates helper in lib file map

richardmilles:
Point contributors at lib/plugin-updates.ts used by /api/plugins/check, next to the existing npx runner note.


agegr (approved):
Approving with a small edit pushed to the branch: #837 (which I've approved) moves the npm/npx CLI lookup into `lib/node-cli.ts`, so the file map now lists that helper and keeps `npx.ts` described as the npx runner rather than repeating the Windows note on it.

PR #823, opened 2026-09-12
fix(cli): disable Wasm lazy compilation on RISC-V

### LINKED ISSUE #685: Crashes with SIGILL on RISC-V (riscv64): next-server dies ~5s after Ready — needs --no-wasm-lazy-compilation
# pi-web crashes with SIGILL on RISC-V (riscv64): next-server dies ~5s after "Ready" — needs `--no-wasm-lazy-compilation`

## Summary

On RISC-V single-board computers (SpacemiT K1 / Lichee Pi 3A in our case), `pi-web` starts normally but the `next-server` child process is killed by **SIGILL (illegal instruction)** about 5–8 seconds after `✓ Ready`, with **no client traffic involved**. The parent monitor reports:

```
▲ Next.js 16.3.1
- Local:         http://127.0.0.1:30141
✓ Ready in 1632ms
⚠ next-swc does not have native bindings support for target triple [object Object]. ...
[pi-web] Next.js exited unexpectedly (signal SIGILL)
```

We traced it to a V8 bug in wasm lazy-compilation on riscv64 (see linked node report below). **A one-line change to the `spawn()` call in `bin/pi-web.js` fully fixes it on our hardware.** Verified stable for 90+ s with real page rendering.

## Environment

| | |
|---|---|
| Board | Sipeed Lichee Pi 3A — SpacemiT K1, 8× T-Heng X60, 16 GB RAM |
| OS | Bianbu 1.0.15 "Mantic Minotaur" (Ubuntu mantic base), kernel 6.1.15, glibc 2.38 |
| CPU ISA | `rv64imafdcv_sscofpmf_sstc_svpbmt_zicbom_zicboz_zicbop_zihintpause` — **no Zba/Zbb/Zcb**, no `riscv_hwprobe` syscall on this kernel |
| Node | v26.0.0, unofficial-builds `linux-riscv64` (non-pointer-compression); also reproduced on v24.20.0 |
| pi-web | 0.8.11 (Next.js 16.3.1) |

## Root cause chain

1. Next.js has no native SWC binding for `linux/riscv64`, so it falls back to the **wasm build of SWC** (`next/wasm/@next/swc-wasm-nodejs/wasm_bg.wasm`). (The loader in `next/dist/build/swc/index.js` also lacks a `riscv64` entry entirely — same failure mode documented here: https://dev.to/gounthar/the-one-line-patch-that-unlocked-nextjs-on-risc-v-a-detective-story-71h)
2. V8's wasm **lazy compilation** is broken on riscv64 in these builds: a JS→wasm call jumps into a zero-filled (never-populated) code page → SIGILL. gdb shows the crash `ra` inside `Builtins_JSToWasmWrapperAsm+184`, crash PC in an anonymous code-space mapping filled with `0x00000000`:
   ```
   Thread 1 "next-server (v1" received signal SIGILL, Illegal instruction.
   => 0x0000003fa86cba28:  unimp   (x/8i: eight consecutive `unimp`)
   x/8xw $pc-16: 0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x0
   ra = Builtins_JSToWasmWrapperAsm + 184
   ```
   Kernel dump (identical in every crash, page offset `0xa28` constant, `t0=0x6bae` constant):
   ```
   next-server (v1[4259]: unhandled signal 4 code 0x1 at 0x00000011d183fa28
   cause: 0000000000000002   (RISC-V: illegal instruction)
   ```
3. Disabling wasm lazy compilation (`--no-wasm-lazy-compilation`, i.e. eager compile at module instantiation) **completely avoids the crash** — everything else in pi-web works.

We verified that the following did **not** help (all still SIGILL): `--no-wasm-code-gc`, `--no-wasm-tier-up`, `--no-liftoff`, `--no-flush-baseline-code --no-flush-liftoff-code --no-flush-bytecode --no-flush-code-based-on-time`, `OPENSSL_riscvcap=` (empty), `NEXT_TELEMETRY_DISABLED=1`, downgrading to Node v24.20.0. Simple hand-written wasm (tiny function ×500k calls, and a 5000-function module) runs fine — the trigger requires the big Rust→wasm module (SWC).

## Suggested fix

In `bin/pi-web.js`, the child spawn currently looks like:

```js
const child = spawn(process.execPath, [nextBin, ...nextArgs], {
  cwd: pkgDir,
  stdio: ["inherit", "pipe", "inherit"],
  env: { ...process.env, PI_WEB_HOSTNAME: hostname },
});
```

Pass the flag as a real argv element (guard on `process.arch`, zero impact on other platforms):

```diff
-const child = spawn(process.execPath, [nextBin, ...nextArgs], {
+const child = spawn(
+  process.execPath,
+  [
+    // riscv64-wasm-lazy-workaround: V8 wasm lazy-compilation on riscv64
+    // jumps into unpopulated code pages -> SIGILL — nodejs/node#65724.
+    ...(process.arch === "riscv64" ? ["--no-wasm-lazy-compilation"] : []),
+    nextBin,
+    ...nextArgs,
+  ],
+  {
   cwd: pkgDir,
   stdio: ["inherit", "pipe", "inherit"],
   env: { ...process.env, PI_WEB_HOSTNAME: hostname },
-});
+  }
+);
```

Notes from our testing:

* `spawn()`'s **`execArgv` option did not reach the child in our tests** — this also holds on official linux-x64 builds, not a riscv64 quirk (nodejs/node#65724, with the spawn() quirk in nodejs/node#65725). Prepending into the argv list is the reliable form.
* Alternatively, an escape hatch env var (e.g. `PI_WEB_NODE_FLAGS`) honored in the spawn would let downstream users work around V8/platform bugs without waiting for releases.

## Reproduction (any riscv64 SBC with ≥4 GB RAM)

```bash
npm i -g @agegr/pi-web@0.8.11
pi-web --port 30141
# dies ~5-8s after "Ready"; monitor prints:
#   [pi-web] Next.js exited unexpectedly (signal SIGILL)
# control:
node --no-wasm-lazy-compilation "$(npm root -g)/@agegr/pi-web/bin/pi-web.js" --port 30141
# stays alive; GET / returns the real page (we served 9756-byte HTML repeatedly for 90s, 0 SIGILL)
```

## Upstream links

* Node/V8 wasm lazy-compilation bug: nodejs/node#65724
* Next.js riscv64 platform detection (missing `riscv64` in `getSupportedArchTriples`): https://github.com/vercel/next.js — related write-up https://dev.to/gounthar/the-one-line-patch-that-unlocked-nextjs-on-risc-v-a-detective-story-71h

— Acidmoon, KUBUDS Tech

huiq777:
## Summary

- start the packaged Next server with `--no-wasm-lazy-compilation` on `riscv64`
- preserve the existing Node argv on other architectures
- cover the architecture branch and caller-owned argument behavior with dependency-free tests

Closes #685

## Why

Node/V8 flags must appear before the Next CLI entrypoint. Scoping the workaround to `process.arch === "riscv64"` avoids changing compilation behavior for x64 and arm64 installs. A small argv helper keeps the architecture branch directly testable without changing the CLI's cwd, environment, stdio, browser launch, or child-process lifecycle.

## Testing

- `env PI_CODING_AGENT_DIR=/private/tmp/pi-web-685-agent npm test` — 1,023 passed
- `npm run lint`
- `node --test lib/pi-web-node-args.test.mjs lib/pi-web-options.test.mjs lib/node-version.test.mjs lib/process-lifecycle.test.mjs` — 26 passed
- `node --check bin/pi-web.js`
- `node --check bin/pi-web-node-args.js`
- `NPM_CONFIG_CACHE=/private/tmp/pi-web-685-npm-cache npm pack --dry-run --ignore-scripts` — packaged helper confirmed
- `git diff --check`

## Evidence boundary

The argv behavior is covered by red/green tests and Node accepts the V8 flag locally. I do not have RISC-V hardware, so the reported target-device `SIGILL` still needs confirmation by the reporter or CI.

### DIFF
--- a/bin/pi-web-node-args.js
+++ b/bin/pi-web-node-args.js
@@ -1,0 +1,11 @@
+"use strict";
+
+function getNextNodeArgs(nextBin, nextArgs, arch = process.arch) {
+  if (arch === "riscv64") {
+    return ["--no-wasm-lazy-compilation", nextBin, ...nextArgs];
+  }
+
+  return [nextBin, ...nextArgs];
+}
+
+module.exports = { getNextNodeArgs };
--- a/bin/pi-web.js
+++ b/bin/pi-web.js
@@ -18,6 +18,8 @@ const fs = require("fs");
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 const { getHelpText, parseLaunchOptions } = require("./pi-web-options");
 // eslint-disable-next-line @typescript-eslint/no-require-imports
+const { getNextNodeArgs } = require("./pi-web-node-args");
+// eslint-disable-next-line @typescript-eslint/no-require-imports
 const { wireChildProcessLifecycle } = require("./process-lifecycle");
 
 let launchOptions;
@@ -81,7 +83,7 @@ nextArgs.push("-H", hostname);
 
 // Always run next's JS entry with node directly — avoids .bin symlink issues
 // and path-with-spaces problems on Windows when shell: true is used.
-const child = spawn(process.execPath, [nextBin, ...nextArgs], {
+const child = spawn(process.execPath, getNextNodeArgs(nextBin, nextArgs), {
   cwd: pkgDir,
   stdio: ["inherit", "pipe", "inherit"],
   env: { ...process.env, PI_WEB_HOSTNAME: hostname },
--- a/lib/pi-web-node-args.test.mjs
+++ b/lib/pi-web-node-args.test.mjs
@@ -1,0 +1,30 @@
+import assert from "node:assert/strict";
+import { createRequire } from "node:module";
+import test from "node:test";
+
+const require = createRequire(import.meta.url);
+const { getNextNodeArgs } = require("../bin/pi-web-node-args.js");
+
+test("passes the no-wasm-lazy-compilation flag to Node on RISC-V", () => {
+  assert.deepEqual(getNextNodeArgs("next-bin", ["start", "-p", "30141"], "riscv64"), [
+    "--no-wasm-lazy-compilation",
+    "next-bin",
+    "start",
+    "-p",
+    "30141",
+  ]);
+});
+
+test("keeps the existing Node arguments on other architectures", () => {
+  for (const arch of ["x64", "arm64"]) {
+    assert.deepEqual(getNextNodeArgs("next-bin", ["start"], arch), ["next-bin", "start"]);
+  }
+});
+
+test("does not mutate the caller's Next.js arguments", () => {
+  const nextArgs = ["start", "-H", "127.0.0.1"];
+
+  getNextNodeArgs("next-bin", nextArgs, "riscv64");
+
+  assert.deepEqual(nextArgs, ["start", "-H", "127.0.0.1"]);
+});


agegr (approved):
Scoped to `riscv64`, flag placed before the Next entrypoint as V8 requires, `bin/` is already in the published `files`. Tests pass, lint clean. Approving; #685's reporter can confirm on hardware once this ships.

PR #825, opened 2026-09-12
feat(sidebar): make conversation and file panes resizable

lyx2145181:
## Summary

- add a draggable horizontal separator between conversations and the file explorer
- persist the conversation pane height in local storage and clamp it to keep both panes usable
- support Arrow Up/Down, Enter, and double-click reset with separator semantics and visible keyboard focus
- preserve resizing while conversation search results are active
- generalize the existing resizable-panel hook for vertical movement without changing current horizontal panel behavior

## Testing

- targeted sidebar and resizable-panel tests: 25 passed
- `npm run lint`
- `git diff --check upstream/main...HEAD`

Manual interaction on the source implementation also verified drag resizing and persistence across reloads.


Light1002:
I would find this useful in my daily Pi + Pi Web workflow on Windows. When browsing agent-generated files, the current non-adjustable vertical split gives the file tree too little space; when reviewing history, I need more room for sessions. Horizontal sidebar resizing does not address this.

This PR appears to cover the behavior I was looking for. In particular, I would appreciate preserving the chosen split after collapsing/reopening the Explorer and after a page reload, while keeping both panes usable on smaller windows.

Please also consider checking that session search, inline session renaming, and Explorer collapse/expand continue to work correctly with the resizable split. I have not tested this PR branch; this is use-case feedback rather than a verification report.

agegr (approved):
Reasonable ergonomics fix (the fixed split does starve the file tree), and it reuses `useResizablePanel` with an `axis` option instead of adding a second hook — good. Keyboard/double-click reset and the clamp against the explorer minimum look right. Merges cleanly onto current `main`; tsc, eslint, and the SessionSidebar / useResizablePanel tests pass.

Approving. One heads-up: the session sidebar is due for a redesign, so merging may wait until that plan is settled; the divider itself is orthogonal to the session list content, so it should carry over.

PR #826, opened 2026-09-12
Show tool-result images while the tool card is collapsed

uvforce:
## Problem

Tool results that include `ImageContent` (browser screenshots, generated images) are only rendered inside the expanded result area. The default collapsed tool card therefore hides the one thing the user usually wants to see.

Assistant markdown images have a second, related gap: user-message and tool-result images already open `ImagePreview` on click, but a markdown `` stays a plain `` and cannot be enlarged.

## Change

- Render the tool-result image grid directly under the tool header. Arguments and text output stay behind the existing expand toggle. No tool-name checks, no new API, no new dependency.
- When a result has images and empty text, expanding no longer shows the "no output" placeholder.
- Edit-tool diffs no longer swallow images; the image grid is independent of the text/diff branch.
- Wrap markdown images in the same official `ImagePreview`. The dialog still mounts only after the click, so server rendering never emits a `<dialog>` inside a `<p>`.

This is content-based: any `ImageContent` block and any markdown image works the same way, regardless of which tool or model produced it.

Collapsed cards with many historical screenshots will keep those images in the DOM (`loading="lazy"` already applies). That is the intended default; a setting can be added later if it becomes a problem.

## Tests

- `shows tool-result images while the tool details stay collapsed`
- `opens markdown images in the shared image preview`
- `npm test` for `MessageView` and `MarkdownBody`: 32 passed
- TypeScript, ESLint, and `git diff --check` passed

### DIFF
--- a/components/ImagePreview.tsx
+++ b/components/ImagePreview.tsx
@@ -13,6 +13,8 @@ interface ImagePreviewProps {
 
 export function ImagePreview({ src, alt = "", children, className, style }: ImagePreviewProps) {
   const { t } = useI18n();
+  const previewLabel = t("chat.previewImage");
+  const triggerLabel = alt.trim() ? `${previewLabel}: ${alt}` : previewLabel;
   const [open, setOpen] = useState(false);
   const triggerRef = useRef<HTMLButtonElement>(null);
   const dialogRef = useRef<HTMLDialogElement>(null);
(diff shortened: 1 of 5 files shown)


uvforce:
Addressed the two Codex notes in de9e760:

- Linked images (`[alt](url)`) stay links. The preview button is skipped when the image is already inside an `<a>`, so a click no longer opens the dialog and navigates at the same time.
- The preview trigger keeps the generic “Preview image” label when alt is empty, and uses `Preview image: {alt}` when markdown provides alt text.

uvforce:
@codex review

agegr (approved):
Content-based, no new API or dependency, and the collapsed card now shows the one thing a screenshot-producing tool exists to show. Splitting `ResultImages` out of `PairedResult` also fixes the edit-diff branch swallowing images. The link-context guard for `[](url)` and the alt-aware trigger label are good follow-ups; markdown images were already `display: block`, so the `ImagePreview` button wrapper does not change layout.

Merges cleanly onto current `main`; tsc, eslint, and the MessageView / MarkdownBody / ImagePreview tests pass. Approving.

PR #827, opened 2026-09-13
fix(plugins): normalize relativePath separators for Windows

cy7372:
Node's `path.relative()` returns backslashes on Windows, so the plugins API leaked platform-specific separators into `relativePath` — the bundled route test (`app/api/plugins/route.test.mjs`) asserts `extensions/rtk.ts` and fails on Windows.

```diff
-  return rel && !rel.startsWith(..) ? rel : resource.path;
+  return rel && !rel.startsWith(..) ? rel.split(sep).join(/) : resource.path;
```

The test now passes on Windows; no behavior change on POSIX (`sep` is `/` there).


agegr (approved):
Correct: `path.relative` yields backslashes on Windows and the route test asserts forward slashes. No-op on POSIX. Approving.

PR #828, opened 2026-09-13
feat(chat): add /auto-compact slash command to toggle auto-compaction

cy7372:
The wrapper state already exposes `autoCompactionEnabled` and the SDK accepts `set_auto_compaction`, but nothing in the UI ever sends it — a session that disabled auto-compaction (e.g. via the TUI or another client) could not be re-enabled from the web chat.

Adds:

- `/auto-compact` slash command: toggles auto-compaction for the active session and reports the new state
- state plumbing: `autoCompactionEnabled` is mirrored from the agent state on load and on reconnect, so the toggle reflects changes made elsewhere
- optional `autoCompactionEnabled` field on the client-side `AgentStateResponse` type

No server changes needed — the route already returns the field.


agegr (approved):
Small and useful — the wrapper already exposed `autoCompactionEnabled` and forwarded `set_auto_compaction`, so this just closes the loop from the chat. Approving, with two fixes pushed to the branch:

- **Stale toggle**: the `case "auto-compact"` closure read `autoCompactionEnabled` but the surrounding `useCallback` did not list it in its deps, so the command always saw the initial `true` and could only ever *disable*. eslint flagged it (`react-hooks/exhaustive-deps`); added the dependency, matching how `isCompacting` is handled.
- **Discoverability**: the command was not in `BUILTIN_SLASH_COMMANDS`, so it never appeared in the `/` palette. Added it with en / zh-CN / zh-TW descriptions.

Merged current `main`; tsc, eslint, and the useAgentSession / ChatInput tests pass.

Issue #829, opened 2026-09-13
Programmatic sub-agent dispatch API — design discussion + two reference implementations


northwang-lucky:
First — huge appreciation for the recent sub-agent wave (`b77a25f` queue, `a31d5c5` resume, `2661247` worktree isolation, `e3fbbf6` extension tool selectors, `e83f4b5` cascade delete, `0ff1138` session-lease keep-alive, …). It clearly covers most of the ground we'd been exploring on our fork, and reading through it changed some of what we're proposing.

**The gap we'd like to discuss.** Everything above is driven through the chat surface (the `Agent` tool + profiles). What's still missing for us is a **programmatic entry point**: an in-process API that lets trusted server-side extensions (our use case: a task-runner adapter living inside the pi-web process) start, steer, and settle sub-agent runs without going through a chat turn.

**Sketch** (as implemented on our branches):

- `registerDispatchRuntime()` exposes a versioned `globalThis.__piSubagentDispatch` with `startSubagentDispatch(parentSessionId, params)`
- per-dispatch controls that layer on top of profiles: `tools` / `disallowedTools`, `extensions` / `denyExtensions` (source-aware filter keys: npm package names, or basename for file-path extensions), `excludeTools` (reserved control names always excluded)
- lifecycle events (`started` -> `completed`/`aborted`) carrying the **authoritative effective values** (model / thinking / active tool set) resolved by the runtime — not echoes of the request
- per-dispatch `persistSession: false` -> child runs on an in-memory `SessionManager` (nothing hits disk), with the run's audit metadata redirected to the parent session
- **deliberately no concurrency policy**: queueing and per-parent limits belong to `SubagentQueue`; we dropped our own cap implementation entirely

**Two reference implementations**, same behavior contract, different structure — happy to rework whichever you'd merge, or close both if this doesn't fit the roadmap:

- incremental: https://github.com/northwang-lucky/pi-web/tree/feat/subagent-dispatch-api (5 commits)
- unified resolution pipeline: https://github.com/northwang-lucky/pi-web/tree/feat/subagent-dispatch-api-pipeline (5 commits)

Both branch off current `main` (`8366762`), share a **byte-identical dispatch contract suite**, and land at 1064 / 1065 tests green (upstream 1020 + ours; `tsc` / `eslint` clean). Validated live against a real model: foreground dispatch round-trip, non-echoed effective values, `persistSession:false` leaves no `.jsonl` with audit on the parent, and the profile-level fallback.

[...]

PR #830, opened 2026-09-13
fix: surface when a response is truncated by the output limit

northwang-lucky:
**The problem (observed in real use).** When an assistant turn ends on `stopReason: "length"` — e.g. a thinking model spending the entire output budget on reasoning — the conversation stops with no visible answer and no warning; the UI just sits on the last thinking block. It reads exactly like a hung session. We hit this with a 128K-token reasoning-only truncation on a long session: the model never produced a final message, and nothing in the UI said so.

**The fix.** Mirrors the existing provider-error handling:

- `getAssistantTruncationNotice()` in `lib/message-display.ts` — a pure function beside `getAssistantErrorMessage()`: non-streaming + `stopReason === "length"` -> a short notice. `stopReason` reaches `MessageView` intact on both the live event path (`normalizeToolCalls` preserves it through the spread) and the history path (verified).
- Rendered as an inline notice beside the existing error banner (`MessageView`), including the thinking-only-truncated case; classified alongside errors in the final-answer logic (`ChatWindow`) so the message is never silently dropped.

Notice copy: *"This response was cut off after reaching the model's output limit. Send a follow-up to continue."*

**Tests.** 5 new cases (3 on the pure function covering the discriminating states, 2 real-render cases on `MessageView` including thinking-only). `tsc` / `eslint` clean; 1025 tests pass on this branch.

Found while building the programmatic dispatch API discussed in #829 — unrelated to that design, so it ships standalone.


agegr (approved):
Real gap: a `stopReason: "length"` turn with reasoning-only output looked exactly like a hung session. Mirroring the provider-error path is the right shape, and the ChatWindow final-answer classification fix means the message is not silently dropped.

One change pushed to the branch: the notice copy was a hardcoded English string in `lib/message-display.ts`; all UI copy here goes through i18n. The helper is now `isAssistantTruncated(message, opts): boolean` and `MessageView` renders `t("chat.truncatedByOutputLimit")` (en / zh-CN / zh-TW added). Tests adjusted; the render tests still assert the English copy through `I18nProvider`.

Merged current `main`; tsc, eslint, and message-display / MessageView / ChatWindow tests pass. Approving.

PR #831, opened 2026-09-13
feat(subagents): programmatic dispatch API with per-dispatch controls (incremental)

northwang-lucky:
Part of the discussion in #829 — this is the **incremental** of the two reference implementations (the pipeline variant is #832). Either one is ready to walk forward; whichever the maintainer prefers, we'll rework against your review (or close both without hard feelings).

**What it adds** (5 commits, branched from `main` @ `8366762`):

- `lib/subagent-dispatch.ts` — a versioned programmatic dispatch API (`registerDispatchRuntime()` -> `globalThis.__piSubagentDispatch`): `startSubagentDispatch(parentSessionId, params)` with per-dispatch `tools` / `disallowedTools`, `extensions` / `denyExtensions` (source-aware filter keys), `excludeTools` (reserved control names always excluded), `signal` / `onUpdate`, and lifecycle events carrying the **runtime-resolved effective model/thinking/activeTools** — not request echoes
- per-dispatch `persistSession: false` — child session runs in-memory (nothing hits disk); the run's audit metadata is redirected to the parent session, reusing the existing `pi-web:subagent` custom-entry format
- deliberately **no** concurrency policy: queueing and per-parent limits stay with the upstream `SubagentQueue` (`b77a25f`)

**Tests**: shared contract suite (`subagent-dispatch*.test.mjs`, `subagent-tools-resolution.test.mjs`) is byte-identical with the pipeline branch; this branch lands at 1064 passing (upstream 1020 + 44), `tsc` / `eslint` clean. Behavior validated live against a real model (dispatch round-trip, non-echoed effective values, in-memory run leaves no `.jsonl`).

**Known coverage note**: the profile-level `persistSession=false` case is pinned by a source-level assertion here; a driven behavioral test for it exists on #831 and we'll port it over if this branch becomes the chosen one.


agegr:
Thank you for the care that went into this (and for offering two variants to pick from). I'm closing both #831 and #832, and the reason is scope rather than quality.

pi-web is a web front-end for pi: it renders sessions, sends prompts, and exposes what the pi SDK already offers. A programmatic dispatch API (`globalThis.__piSubagentDispatch`, per-dispatch tool/extension filters, in-memory child sessions with redirected audit metadata, lifecycle event contracts) is a platform/orchestration layer. Nothing in the web UI needs it, and once it exists pi-web becomes responsible for a public API surface that other code depends on — versioning, compatibility, and semantics that belong in pi itself or in an extension, not in the web wrapper. ~1700 lines per variant is also more than I want this repo to carry for a feature the UI does not consume.

If the underlying need is "let an extension dispatch sub-agents with per-run controls", the pi extension API / upstream SDK is where I'd raise it. Happy to keep the discussion open on #829 for anything that does surface in the web UI.

PR #832, opened 2026-09-13
feat(subagents): programmatic dispatch API via a unified resolution pipeline

northwang-lucky:
Part of the discussion in #829 — this is the **unified resolution pipeline** of the two reference implementations (the incremental variant is #831). Same behavior contract (the dispatch contract suite is byte-identical between the two branches), different internal structure: one `resolveSubagentResources()` pipeline produces a typed resolution plan consumed once by the runtime, instead of per-control inline logic. Either one is ready to walk forward; whichever the maintainer prefers, we'll rework against your review (or close both without hard feelings).

**What it adds** (5 commits, branched from `main` @ `8366762`):

- `lib/subagent-dispatch.ts` — programmatic dispatch API (`registerDispatchRuntime()` -> `globalThis.__piSubagentDispatch`: `startSubagentDispatch(parentSessionId, params)`, per-dispatch `tools` / `disallowedTools` / `extensions` / `denyExtensions` / `excludeTools`, `signal` / `onUpdate`, lifecycle events carrying runtime-resolved effective model/thinking/activeTools — not request echoes) plus the unified `resolveSubagentResources()` plan
- `lib/subagent-state.ts` — one typed `globalThis` registry replacing scattered lazy-global writes, hot-reload safe
- per-dispatch `persistSession: false` — in-memory child session (nothing hits disk) with audit metadata redirected to the parent; profile-level fallback additionally pinned by a driven behavioral test
- no concurrency policy: queueing and per-parent limits stay with the upstream `SubagentQueue` (`b77a25f`)

**Tests**: 1065 passing (upstream 1020 + 45), `tsc` / `eslint` clean; same live validation as #831 (dispatch round-trip, non-echoed effective values, in-memory run leaves no `.jsonl` with audit on the parent).


agegr:
Thank you for the care that went into this (and for offering two variants to pick from). I'm closing both #831 and #832, and the reason is scope rather than quality.

pi-web is a web front-end for pi: it renders sessions, sends prompts, and exposes what the pi SDK already offers. A programmatic dispatch API (`globalThis.__piSubagentDispatch`, per-dispatch tool/extension filters, in-memory child sessions with redirected audit metadata, lifecycle event contracts) is a platform/orchestration layer. Nothing in the web UI needs it, and once it exists pi-web becomes responsible for a public API surface that other code depends on — versioning, compatibility, and semantics that belong in pi itself or in an extension, not in the web wrapper. ~1700 lines per variant is also more than I want this repo to carry for a feature the UI does not consume.

If the underlying need is "let an extension dispatch sub-agents with per-run controls", the pi extension API / upstream SDK is where I'd raise it. Happy to keep the discussion open on #829 for anything that does surface in the web UI.

Issue #804, opened 2026-09-11
Bug: "Chat only" tool preset drops extension-registered model providers, so sending fails with "Model not found"


612cs:
### Summary

With the composer tool preset set to **Chat only** (`none`), a session never loads extensions, so any provider registered by an extension package is missing from that session's `modelRuntime`. `/api/models` — which populates the composer's model selector — always loads extensions, so it still reports such a model as the session default. Pressing Send then fails with a bare `Model not found: <provider>/<modelId>` notice and the message is never sent.

This is not a network or provider problem. The same model works fine in a `default` preset session, and `/v1/models` on the provider responds normally.

### Environment

- pi-web 0.9.1 (global install), Node 22.22.2, macOS
- Provider comes from an extension package: `npm:@router-for-me/pi-cliproxyapi-provider` v1.4.15, enabled via `packages` in `~/.pi/agent/settings.json`
- `~/.pi/agent/auth.json` is empty and there is no `~/.pi/agent/models.json`, i.e. **the extension-registered provider is the only model source**. Any session that skips extensions therefore has zero usable models.

### Reproduction

1. Install a provider extension (e.g. `@router-for-me/pi-cliproxyapi-provider`) and configure it.
2. New session, then set the composer tool preset to **Chat only**.
3. Press Send (or change the model in the selector).

Same thing at the HTTP level:

```bash
# Chat only (toolNames: []) -> provider never registered, session gets the placeholder model
POST /api/agent/new {"cwd":"...","type":"ensure_session","toolNames":[]}
# -> {"success":true,"model":{"provider":"unknown","modelId":"unknown"},"thinkingLevel":"off"}

POST /api/agent/<id> {"type":"set_model","provider":"cliproxyapi","modelId":"gemini-3.8-flash-high"}
# -> {"error":"Model not found: cliproxyapi/gemini-3.8-flash-high"}

# control: default preset, same cwd
POST /api/agent/new {"cwd":"...","type":"ensure_session","toolNames":["read","bash","edit","write"]}
# -> {"success":true,"model":{"provider":"cliproxyapi","modelId":"gemini-3.8-flash-high"},"thinkingLevel":"low"}

POST /api/agent/<id2> {"type":"set_model","provider":"cliproxyapi","modelId":"gemini-3.8-flash-high"}
# -> {"success":true,"data":{"id":"gemini-3.8-flash-high","provider":"cliproxyapi"}}
```

And the selector data is produced independently of the chat-only setting:

```
[...]

Issue #309, opened 2026-07-30
Bug: Anthropic (and other dual-auth providers) never appear in the Models config panel


opsCar:
## Summary

The Models panel (`ModelsConfig`, `~/.pi/agent/models.json`) never lists `anthropic`, even when Anthropic is fully configured and its models are usable in the chat model selector. The same gap applies to `github-copilot` and `openai-codex`. Only custom `models.json` providers show up, so a working provider looks unconfigured and its credential cannot be inspected, replaced, or removed from the UI.

## Reproduce

1. Store an Anthropic API key (via `pi` CLI, or `ANTHROPIC_API_KEY` → `pi /login`), so `~/.pi/agent/auth.json` contains:
   ```json
   { "anthropic": { "type": "api_key", "key": "sk-ant-…" } }
   ```
2. Add one or more custom providers in `~/.pi/agent/models.json` (e.g. `acme-gateway`, `acme-gateway-chat`, `acme-gateway-openai`).
3. Open Pi Web → sidebar → **Models**.

**Expected:** the provider list shows Anthropic (with its auth status) alongside the custom gateway providers.
**Actual:** the list contains only the three custom gateway providers. Anthropic is absent, while its models are selectable in the chat model dropdown — the two views disagree about which providers exist.

## Verified root cause

Anthropic falls through a gap between the two provider-listing endpoints:

- `app/api/auth/all-providers/route.ts` (API-key providers → `activeApiKey`):
  ```ts
  const OAUTH_PROVIDER_IDS = new Set(["anthropic", "github-copilot", "openai-codex"]);
  ...
  if (OAUTH_PROVIDER_IDS.has(provider.id) || !provider.auth.apiKey?.login) continue;
  ```
  Anthropic is skipped as "OAuth-only, handled separately".

- `app/api/auth/providers/route.ts` (OAuth providers → `activeOAuth`):
  ```ts
  const EXCLUDED = new Set(["anthropic"]);
  ```
  Anthropic is skipped here too, so nothing handles it.

`ModelsConfig.tsx` renders exactly `activeOAuth` + `activeApiKey` + `models.json` providers, and the "Add provider" picker uses the same two lists (`availableOAuth` / `availableApiKey`), so Anthropic is unreachable in both places.

Runtime facts from a real config (provider names redacted):

```
anthropic auth methods : { apiKey: "Anthropic API key", oauth: "Anthropic (Claude Pro/Max)" }
getProviderAuthStatus  : { configured: true, source: "stored" }
[...]

PR #833, opened 2026-09-13
fix(models): include extension-registered providers in settings and auth routes

elrond298:
## Problem

Providers registered by an extension package (any extension that calls
`pi.registerProvider` / `createProvider` while loading resources) never appear in the
Models config panel, so their credentials cannot be managed from Pi Web — even though
the composer's model selector offers them and they work in a session.

With such an extension installed and its key stored in `auth.json`:

- `/api/models` → lists the provider's models ✅
- `/api/auth/providers` → `apiKeyProviders` has no entry for that provider ❌

## Root cause

Extension providers reach a model runtime in `createAgentSessionServices()`: it runs the
resource loader, then replays `pendingProviderRegistrations` onto its `ModelRuntime`
(`dist/core/agent-session-services.js`).

`/api/models` builds its runtime that way — hence the model is offered in the composer.
The four auth routes instead called a bare `ModelRuntime.create()`, which only knows
built-in providers plus `models.json`. `getProviders()` never contained the extension
provider (so the panel could not list it) and `getProvider(id)` returned `undefined` for
its login / logout / save-key paths.

## Fix

New `lib/model-runtime.ts` exports `createModelRuntimeWithExtensions()`, which builds the
runtime through `createAgentSessionServices({ cwd: agentDir, agentDir })`. All four auth
routes now use it:

- `app/api/auth/providers/route.ts` — panel listing
- `app/api/auth/api-key/[provider]/route.ts` — save key
- `app/api/auth/login/[provider]/route.ts` — OAuth / device-code login
- `app/api/auth/logout/[provider]/route.ts` — OAuth logout

Deliberate choices:

- **cwd is the agent dir, not a request-supplied project.** Only global package extensions
  load; a browsed repository's project-local `.pi/extensions` are never imported, so these
  routes cannot become a new execution path for untrusted repository code (related: #236).
- **Not cached.** These routes serve auth status and login/logout, so they need current
  credentials rather than a stale snapshot.

## Verification

Dev server started from this branch (pi-web 0.9.1) with a provider extension installed
(14 models registered, key already in `auth.json`):

```
GET /api/auth/providers
-  before: apiKeyProviders ... "newapi", "oc-sdk-go"        (extension provider absent)
+  after:  apiKeyProviders ..., "<ext-provider>", "newapi", "oc-sdk-go"
+
+  {"id":"<ext-provider>","displayName":"<Ext Provider>","configured":true,
+   "source":"stored","modelCount":14,"supportsOAuth":false}
```

- `/api/models` unchanged — same list, extension models still present
- `POST /api/auth/api-key/<unknown>` still fails cleanly (`... does not support API key
  login`) rather than crashing
- `npm run lint` clean; no type errors in the changed files
- cost: 3.3 s on the first request (Turbopack compile + extension load), 0.4 s steady state

## Scope

- Fixes the settings-panel / credential-management gap only. Related but **not** fixed:
  #804 — a "Chat only" session never loads extensions, so that session's own runtime
  still lacks extension providers. That path is session creation (`startRpcSession`),
  not the auth routes.
- `lib/provider-usage.ts` contains the same bare `ModelRuntime.create()`; quota lookup
  for extension providers stays unavailable there. Left out to keep this diff small.
- Same panel as #309 (dual-auth providers), different cause.


agegr (approved):
Confirmed the asymmetry: `/api/models` builds its runtime through `createAgentSessionServices()` (so extension providers show up in the composer) while the four auth routes used a bare `ModelRuntime.create()`. Routing them through one helper that loads only global extensions (agent dir as cwd, so untrusted project extensions stay out) is the right fix and matches the existing pattern. Merges cleanly onto `main`; tsc, eslint, and the auth / provider-listing tests pass. Approving.

PR #835, opened 2026-09-13
fix(streaming): stop duplicating the first streamed chunk

grubwithu:
Pi's assistant `message_start` snapshot can already contain a block's first delta: pi-ai documents `partial` as a shared live response-so-far helper, not an event-time snapshot, so a snapshot read at `*_start` may carry the chunk that the following `*_delta` delivers again.

The reducer kept that text when `text_start`/`thinking_start` arrived and then appended the delta a second time, rendering the first word twice until the authoritative `*_end` replaced the whole block. Streamed tool input had the same problem.

Reset text, thinking, and tool-call blocks on their `*_start` event instead, matching pi-ai's documented semantics (empty at start, grown only by deltas, `*_end` authoritative) and pi's own AssistantMessageEventProxy.

Adds regression coverage for the leaked-first-chunk sequence in lib/streaming-message.test.mjs.

Before fix:
8bcbea0c99ad692dce56fb895f98cb64

After fix:
5dfb46ab71213caaa124afe1b7813f4b


agegr (approved):
Matches pi-ai's documented contract (`partial` is the shared live helper; blocks are empty at `*_start`, grown by `*_delta`, `*_end` authoritative), and the screenshot shows exactly the doubled first word. Resetting on `*_start` is the correct reducer behavior; the `*_end` replacement still covers redacted thinking and provider-specific initial tool args. Regression test included; merges cleanly, tests and lint pass. Approving.

PR #836, opened 2026-09-13
feat(chat): load earlier history from the minimap

grubwithu:
The minimap only indexed the loaded page, and older pages were fetched solely by the top sentinel when the main chat area scrolled up. A single agent turn can span hundreds of entries, so the default 50-entry page often holds zero or one user turn and the minimap loses its navigation value.

Add a "Load earlier" row at the top of the minimap preview. Extract the sentinel's fetch into one guarded loadOlderPage shared by both triggers, so the row reuses the in-flight guard and scroll anchoring. The preview panel now also renders when no turn is loaded yet but earlier history exists, which is exactly the huge-single-turn case.

Verified in a 1428-entry session: 28 loads take the minimap from 1 to all 19 user turns, after which the row disappears; scrolling to the top still paginates through the sentinel.


agegr:
Thanks — the diagnosis is right, but I'd rather fix the cause than add a second pagination trigger. The reason the minimap "loses its navigation value" is that the tail page was counted in raw entries, so tool-heavy turns crowded out user messages. #810 (approved) changes the page budget to count only visible user/assistant messages, so the default page now carries a normal number of user turns and the minimap indexes them without extra loads. The existing top sentinel still paginates for anything older.

With that in place a "Load earlier" row in the minimap is extra UI for a case that no longer needs it, so I'm closing this. If, after #810 lands, there is still a concrete session where the minimap is unusable, please open an issue with the entry mix and I'll look again.

PR #837, opened 2026-09-13
fix(plugins): run npm update checks without the npm.cmd shim on Windows

userliwq:
## Problem

`POST /api/plugins/check` fails with `error spawn npm ENOENT` for every npm-installed plugin on Windows, so the plugin page can never offer an update for them. Git plugins keep working, which makes the failure look source-specific rather than platform-specific:

```
git:github.com/NVlabs/SoL-Pi        git  up-to-date
@ff-labs/pi-fff                     npm  error spawn npm ENOENT
pi-web-access                       npm  error spawn npm ENOENT
rpiv-todo                           npm  error spawn npm ENOENT
@vanillagreen/pi-background-tasks   npm  error spawn npm ENOENT
```

`lib/plugin-updates.ts` spawns the configured npm command with `execFile`. On Windows that command is `npm.cmd`, and Node refuses to spawn `.cmd`/`.bat` files without a shell (CVE-2024-27980), so the call dies before it ever reaches the registry. CI only runs `ubuntu-latest`, so this never surfaced.

## Fix

Resolve `npm`/`npx` through the CLI script bundled with the running Node.js (`node_modules/npm/bin/<name>-cli.js`, or `../lib/node_modules/...` on Unix) and spawn it with the `node` binary: no shell is involved and the behavior is identical on every platform. This is the approach the skills routes already use for npx, so that lookup moves into `lib/node-cli.ts` and `lib/npx.ts` now shares it instead of carrying its own copy. When no bundled CLI script is found, the bare command name is used exactly as before.

A `npmCommand` configured in settings that points at something else is still passed through untouched.

## Verification

- Windows 11, Node 24.20.0, npm 11.19.0: `checkPluginUpdates` with the real runner (no injected command stub) now resolves all five configured plugins, returning `up-to-date` for the four npm ones instead of `error spawn npm ENOENT`.
- `tsc --noEmit` and `npm run lint`: clean.
- `npm test`: 1005 passing, 14 failing — the identical 14 failures occur on unmodified `main` in the same checkout (terminal, file-viewer, PATH and symlink tests that are unrelated to this change), so no new failure is introduced and none is masked.
- `lib/node-cli.test.mjs` covers both install layouts, the probe order and the fallback path.

Note for reviewers: the npx path is behavior-preserving (same candidate list, same order, same fallback), so the skills routes are unaffected.


agegr (approved):
Confirmed #742: `execFile("npm", …)` can never spawn `npm.cmd` on Windows without a shell. Extracting the finder that `lib/npx.ts` already had into `lib/node-cli.ts` and reusing it for `npm view` is the right shape — one implementation, injectable for tests, no shell. Merges cleanly onto current `main`; tsc, eslint, and `lib/node-cli.test.mjs` pass.

Approving. (#817 addressed the same issue with a second copy of the finder; closing that one in favor of this.)

PR #839, opened 2026-09-13
fix: preserve extension widget order on updates

darthrick0:
## Problem

This was observed in Firefox on Windows with the `@tintinweb/pi-subagents` plugin. The independently updating Agents and Fleet widgets repeatedly swapped positions in the bottom widget strip, causing visible flickering.

`setWidget` removed the existing keyed widget and appended its replacement. Consequently, every content update moved that widget to the end of the list. The issue is in the client-side state logic and is not expected to be browser-specific.

## Changes

- Update existing widgets at their current array position.
- Continue appending new widgets and removing explicitly cleared widgets.
- Extract the state update into a small pure helper.
- Add regression coverage for independently updating Agents and Fleet widgets.

## Verification

- `npm test` — 1,022 tests passed
- `npm run lint`


agegr (approved):
Confirmed: the client reducer removed-then-appended on every `setWidget`, so two independently refreshing widgets kept swapping places. Updating in place matches the server side, which keeps widgets in an insertion-ordered `Map` and `set()`s existing keys, so the strip now agrees with what a reconnect would return. Pure helper with a regression test; merges cleanly, tsc/eslint/tests pass. Approving.

PR #845, opened 2026-09-14
feat: scroll to latest button

grubwithu:
## Add a jump-to-latest button to the chat

**Problem:** The right-side minimap can navigate around a session, but once you scroll up there is no quick way back to the live tail — you have to drag the minimap or scroll manually.

**Change:** A circular down-arrow button floats just above the composer whenever the chat viewport is detached from the bottom, and hides again once you are at the tail. Clicking it smooth-scrolls to the latest message.

图片

### Details

- **Visibility** is a pure predicate, `shouldShowScrollToLatest(scrollTop, clientHeight, scrollHeight)` in `lib/chat-lazy-load.ts`. It is intentionally separate from the existing live-follow attachment heuristic, which has its own reattach tolerance semantics — reusing it would have made the button appear and disappear at slightly the wrong moments.
- **Placement** is centered on the chat column and stops short of the minimap (`right: CHAT_MINIMAP_WIDTH`), so it never covers it.
- **Transition** is deliberately understated: a 160ms fade with a 4px rise and a slight scale. The button stays mounted and toggles an `is-visible` class rather than being conditionally rendered, so the *exit* animates too. A delayed `visibility` switch removes it from the tab order once faded. Under `prefers-reduced-motion: reduce`, the fade is kept and the transform is dropped.

### Testing

- New unit coverage for the visibility predicate (`lib/chat-lazy-load.test.mjs`) and for the component wiring/CSS (`components/ChatWindow.scroll-to-latest.test.mjs`).
- Verified in a real browser against a 400-message session: hidden at the tail, appears on scroll-up without overlapping the composer or minimap, and returns to the bottom (then hides) on click. The transition was checked mid-flight to confirm the opacity is genuinely interpolating rather than snapping.
- `tsc --noEmit`, `eslint .`, and the full unit suite (1026 tests) all pass.

No behavior change when the chat is already at the bottom, for empty new sessions, or on mobile.


agegr (approved):
Standard chat affordance, kept small: a pure visibility predicate, one state flag set from the existing scroll handler, and a button that reuses `scrollToBottom`. Keeping it separate from the live-follow reattach heuristic is the right call, and the visibility/pointer-events handling in CSS means it never intercepts clicks while hidden. Merges cleanly; tsc, eslint, and the new tests pass. Approving.

PR #847, opened 2026-09-14
Include the subagent session ID in foreground completion text

uvforce:
## Problem

The `Agent` tool exposes `resume` ("Existing subagent session ID to continue instead of creating a new session") and `get_subagent_result(agent_id)`, both keyed by the subagent session ID. But when a foreground subagent completes, the tool result text is only the subagent's report; the session ID is present only in `details`, which the model never sees. Background runs already include `Session ID: …` in their text, and every other status branch (`running`, `aborted`, `interrupted`, `failed`) includes the ID too.

In practice this makes `resume` unreachable for foreground runs. Observed with a builder profile that has `max_turns`: it hit the limit and handed back a partial report, the parent model tried to `resume` as instructed, guessed three wrong IDs (`Subagent not found`), and had to start a fresh session and redo the work.

## Change

`subagentFinalText` now prefixes the completed text with `Subagent <id> completed.` (and `Subagent <id> completed without text output.` when the run produced no text), matching the format of the other status branches. The report body is unchanged. `details` is untouched.

This also applies to the background completion notification delivered via `notifyParent`, so both paths now consistently carry the ID.

## Verification

- `lib/subagent-extension.test.mjs`: foreground completion, `get_subagent_result` (with and without text), and `resume` now assert the ID-prefixed text.
- `lib/subagent-runtime.test.mjs`: background completion notification asserts the ID-prefixed text.
- `npm test` (38 in the touched suites; full suite 1094 on the fork branch), `tsc --noEmit` clean.


agegr (approved):
Right call: the model only sees `content`, and every other status branch already carries the session ID, so `completed` was the one path where `resume` / `get_subagent_result` had nothing to work from. Prefix format matches the existing branches; tests updated; merges cleanly and lint passes. Approving.

PR #849, opened 2026-09-14
feat(models): manage model visibility from the chat selector

AKAZIK-py:
## Problem

Two long-standing gaps around the model selector:

1. **Deleting a model in Settings → Models doesn't remove it from the chat selector.** `models.json` is an upsert/override layer on top of pi's built-in provider catalogs, so a deleted entry just falls back to the built-in definition for any configured provider (#560, #640). PR #561's implicit prune was rejected as by-design, and the supported lever — the `enabledModels` scope in `~/.pi/agent/settings.json` — had no UI in pi-web at all.
2. **The selector shows every available model with no way to hide any** (#241, #324 — the filter-only half was already shipped). pi's own answer to this is the global `enabledModels` scope, manageable in the TUI via `/scoped-models` but not from the web UI.

## Solution

A **"Manage visible models"** entry at the bottom of the chat model selector dropdown opens a dialog that edits pi's native `enabledModels` scope explicitly:

- Every available model listed, grouped by provider, with filter, per-provider tri-state toggles, and a global select-all
- Checkbox state starts from the models the selector currently shows (the server-resolved scope — no client-side reimplementation of pi's matching rules)
- Saving writes exact `provider/modelId` patterns; checking **every** model clears the scope back to the show-all default, so newly added models keep appearing automatically
- Scopes using globs, bare ids, or `:thinkingLevel` pins are flagged in the dialog before saving replaces them with the explicit list

### API

- `GET /api/models-visibility` — raw `enabledModels` patterns + the full unfiltered catalog (honors project trust like `/api/models`)
- `PUT /api/models-visibility` — validated write (`null` clears), invalidates the models cache and is followed by a `modelsRefreshKey` bump so the selector reloads immediately (#822's refresh path)

### Safety rails

- Empty selection rejected client-side and server-side: pi falls back to showing *every* model when a scope matches nothing, so saving an empty scope would silently undo the edit
- The PUT route resolves the scope before writing and refuses zero-match patterns (a stale/typo'd list can never wipe the selector)

## Relation to #727

Complementary rather than overlapping: #727 adds per-provider catalog cards inside Settings → Models with segment-preserving rewrites; this PR is a lighter, in-flow entry point that manages the whole scope in one view. Both edit the same pi-native setting, so they compose. Happy to rework if you'd rather land only one approach — the API safety rails here (zero-match refusal) and the dialog component are reusable either way.

## Verification

- `npm test` — 1032 pass (incl. new `lib/model-visibility.test.mjs` and `components/ModelVisibilityDialog.test.mjs`)
- `tsc --noEmit`, `eslint`, `next build` clean
- Manual end-to-end against `next dev`: GET/PUT round-trip writes `enabledModels` to settings.json, `/api/models` reflects the scope immediately (cache invalidation), zero-match and empty-array PUTs return 400, `null` restores the file byte-identically

Fixes the hiding half of #241 / #324; gives #560 / #640 users an in-app escape hatch instead of hand-editing settings.json.


agegr (approved):
This is the right lever — editing pi's own `enabledModels` (with the SDK resolver for the current scope, the empty-scope guard, and the advanced-pattern warning) rather than a pi-web-side filter — and it closes the loop on #241 / #324 / #560 / #640. Approving after one structural change, pushed to the branch:

**Entry point moved from the chat model selector to Settings → Models.** Model visibility is configuration, and pi-web keeps configuration in Settings so the composer toolbar stays quiet. Concretely:

- `ModelsConfig` gets a "Manage visible models" action under "Add provider" and hosts `ModelVisibilityDialog`; `SettingsPanel` passes its `cwd` through.
- The `onManage` entry in `ModelSelector`, the `onManageModels` prop on `ChatInput`, and the dialog hosting / `onModelsVisibilityChanged` plumbing in `ChatWindow` and `AppShell` are reverted — closing Settings already bumps `modelsRefreshKey`, so the selector refreshes without new wiring.
- The dialog no longer takes `visibleModels` from the chat; `GET /api/models-visibility` now returns `visible` (resolved server-side via `resolveVisibleModels`) and the dialog seeds its checkboxes from that. Same matcher, one source of truth.
- Without a `cwd`, the route falls back to the agent dir (global extensions only) instead of `process.cwd()`, which would have failed the allow-list check.

Tests updated for the new host; tsc, eslint, and the ModelVisibilityDialog / model-visibility / ModelsConfig / SettingsPanel tests pass on top of current `main`.

agegr:
#930

PR #852, opened 2026-09-14
chore(docs): commit next dev agent rules block to keep tree clean

Zhangs-11:
The AGENTS.md agent-rules block is auto-written and re-added by `next dev` (see `node_modules/next/dist/server/lib/generate-agent-files.js`). Committing it keeps the working tree clean instead of repeatedly showing an unstaged diff.


agegr:
Thanks, but this goes against the convention documented in the same file: `AGENTS.md` asks that the `BEGIN:nextjs-agent-rules` block be treated as generated tooling output and kept out of commits. The block's wording and presence depend on the installed `next` version, so committing it just moves the churn into the history (it would show up as a diff again on the next Next upgrade, and for contributors running a different version). Leaving it untracked and ignoring it in `git status` is the intended workflow. Closing.

PR #857, opened 2026-09-15
fix: stop truncating or dropping select prompt bodies

Nyarlathoteppppp:
## Summary
- Bug: for extension `select()` dialogs, the prompt body was truncated or dropped. Pi Web treated the entire `request.title` as a header, so multiline content never appeared as the dialog body.
- Cause: `select(title, options)` has no separate message field. Plugins such as pi-goal-x put confirmation details in a multiline title (`Confirm Goal Draft\n\nObjective: ...`), which is valid API use.
- Fix: split only on newlines. The first non-empty line stays the heading; the remainder is a scrollable Markdown body. Long single-line titles are unchanged.

This is not specific to pi-goal-x; it is a general rendering gap for `select(title, options)` when the title includes a body. Related to #462 (overlay/layout), but this is a different bug.

## Test plan
- [x] `node --experimental-strip-types --test components/ChatWindow.extension-request.test.mjs lib/extension-dialog-copy.test.mjs`
- [x] `"Choose a model"` stays header-only
- [x] `"Confirm Goal Draft\n\nObjective..."` shows the full body
- [x] A long single-line title stays in the header (no 80-character split)
- [x] Dialog body still uses `minHeight: 0` / `overflowY: auto` so long prompts scroll instead of overflowing


Nyarlathoteppppp:
Closing in favor of fcd94bf, which already preserves multiline select titles in the header like the Pi TUI. Thanks for landing the fix.

PR #861, opened 2026-09-15
feat: one-click 60-second shutdown button with countdown and cancel

NewBugMakerHREAT:
## Summary

Adds a "shut down in 60 seconds" button to the top toolbar. Clicking it opens a confirmation dialog, then starts a 60-second countdown; the machine powers off when it reaches zero. The countdown can be cancelled at any time and survives a server restart.

## Motivation

Pi Web is often used to drive a machine remotely (LAN / phone). A one-click shutdown with a safety net — a confirmation plus a cancellable countdown — lets a user power the machine off from the browser without walking back to it.

## Behavior

State machine: `idle` → (confirm) → `counting` → `done`; `cancel` returns to `idle`.

- Toolbar button next to the existing actions (icon + label on desktop, icon + countdown badge on mobile).
- Confirmation dialog before arming.
- Red countdown `Cancel · Ns left`; click to cancel.
- `GET /api/shutdown` → `{ state, remainingSeconds, deadline, supported }`; `POST /api/shutdown { action: "start" | "cancel" }`.
- Guarded by the existing `proxy.ts` middleware (host allow-list + optional `PI_WEB_PASSWORD` auth) — no new auth code.

## Design notes

- **The OS holds the real timer.** The server runs `shutdown /s /t 60` and only tracks the deadline for UI, so a server crash does not cancel the shutdown.
- **Restart recovery.** The deadline is persisted atomically to `~/.pi/pi-web-shutdown-state.json`. On startup (`instrumentation.ts`) the server re-arms the OS timer with the remaining seconds. An already-expired deadline is dropped — never a surprise shutdown on boot.
- **Cancel is honest.** If `shutdown /a` fails (the OS timer may still be armed), the state stays `counting` and the UI keeps surfacing the failure.
- **Shell-safe.** Commands run through `execFile` with an argument array; nothing reaches a shell.
- **Windows-only for now.** The feature relies on `shutdown /s /t <seconds>`, which has no direct equivalent elsewhere. On non-Windows the route reports `supported: false`, the button is hidden, and start/cancel are rejected with `400`. Cross-platform support (Linux/macOS) can be added later behind a platform-specific command map.

## i18n

Added `shutdown.*` keys to `en`, `zh-CN`, and `zh-TW`.

## Testing

- Unit tests: `lib/shutdown-timer.test.mjs` (10 cases) cover start/cancel/restore, idempotency, rollback on exec failure, and deadline-expiry reconciliation, using injected fakes so no real shutdown is ever executed.
- `tsc --noEmit` and `eslint` pass.
- No e2e test: an end-to-end test would require actually powering off the CI machine.


agegr:
Thanks for the thorough implementation, but this is out of scope for pi-web and I'm closing it.

pi-web is a web front-end for the pi coding agent. Powering off the host machine is a system-administration action that has nothing to do with running an agent session, and putting a `shutdown /s` / `shutdown -h` trigger behind a web endpoint — even with the confirmation dialog, countdown, and the existing password gate — widens what a compromised or mis-shared pi-web URL can do from "read/write files in allowed roots" to "turn the machine off". That is not a trade I want the project to make for a convenience feature; anyone who needs remote power control has purpose-built tools for it.

PR #862, opened 2026-09-15
feat: session timing attribution and an event trajectory panel

612cs:
## What this adds

Two related things about *how long a session actually took*, both of which are currently invisible in the web UI:

1. **Session timing attribution.** `Session info` now answers "where did the wall-clock time go?" — split into model time, tool time, and other. Alongside it, per-request **TTFT** (time to first token) and **output speed** (tokens/s).
2. **An event trajectory panel.** A per-turn ledger of everything that happened in the session — assistant turns, tool calls, compactions, branch summaries — laid out against a timeline so you can see which part of a turn was waiting and which part was decoding.

Sessions accumulate hours of wall-clock time, and the only *time* number the app showed was a single "active" total. That total cannot tell you whether a slow turn was the model thinking, a tool running, or a provider stalled before its first token.

## Timing attribution

`lib/session-timing.ts` walks the same entry-anchor chain `computeSessionTotalActiveMs` already used, so the parts always add up to that total:

```
modelMs + toolMs + otherMs === computeSessionTotalActiveMs()
```

- assistant entry gap -> `modelMs`
- toolResult entry gap -> `toolMs`
- compaction / branch summary -> `modelMs`
- injected messages and unknown types -> `otherMs`

`custom` entries are deliberately excluded from the anchor chain: instrumentation entries are written into the transcript, and letting them move the anchor would make the meter measure itself.

## TTFT and TPS

These cannot be reconstructed after the fact — the session file stores token counts but no timestamps for the first token. So they are captured live by a small built-in extension (`lib/timing-extension.ts`) that observes `before_provider_request`, `message_start`, `message_update`, and `message_end`, and writes the samples back as `pi-web-timing` entries.

- **TTFT** = first `text_delta` / `thinking_delta` / `toolcall_delta` with a non-empty delta.
- **TPS** = `sum(outputTokens) / sum(decodeMs)`, i.e. a weighted average rather than a per-request average, so a fast trivial request cannot flatter the number.
- A request that was aborted or errored keeps its `ttftMs` and zeroes `decodeMs`/`outputTokens`.
- Samples carry a correlation key (message timestamp + response id) so a live sample can be joined to the request it belongs to.

Timing is only available for requests made while this build is running; older sessions report `timedRequests < totalRequests` and the UI says so instead of showing a wrong number.

## Trajectory panel

`lib/trajectory.ts` is pure logic (no React, no IO): it takes session entries plus timing samples and produces turns, rows, and a windowed view. `GET /api/sessions/[id]/trajectory` serves it as a separate route rather than growing the hot-path session payload, and supports `?tail` / `?before` / `?leafId` for paging and for following the selected branch.

The panel shows a sticky time overview (first-token wait / decode / overhead / tool) and a turn-grouped ledger where a row expands into an inspector with model, stop reason, token breakdown, cost, and timing.

One deliberate rule: **segments never draw outside the measured row span.** When a row's reported `durationMs` is shorter than `ttft + decode`, decode is clamped so the bar cannot imply time that was not measured.

V1 intentionally omits wheel zoom, brush selection, drag pan, collapse, search, cross-branch view, and virtualization.

## Verification

- `npm test`: 1059 pass / 0 fail (39 added over main's 1020; new `timing-extension.test.mjs` and `trajectory.test.mjs`, plus added cases in `session-timing.test.mjs`).
- `npx tsc --noEmit` and `npm run lint`: clean.
- `npm run test:e2e`: 7/7 checks pass (against a development server, desktop 1280px and mobile 390px).
- Cross-checked against a real session: `modelMs 701779 + toolMs 519793 = totalActiveMs 1221572` exactly.
- Rendered in a browser at desktop and mobile widths; a mobile layout bug was found and fixed during development (the row grid collapsed to 3 columns while 4 children remained, wrapping the duration onto a second line).

## Files

```
new:
  lib/timing-extension.ts               live TTFT/TPS capture
  lib/timing-extension.test.mjs
  lib/trajectory.ts                     turns, rows, paging, windows (pure logic)
  lib/trajectory.test.mjs
  components/TrajectoryPanel.tsx
  app/api/sessions/[id]/trajectory/route.ts

changed:
  lib/session-timing.ts                 attribution rules, alongside the existing active-time total
  lib/session-timing.test.mjs
  app/api/sessions/[id]/route.ts        response gains `timing`
  hooks/useAgentSession.ts              SessionData.timing
  lib/pi-types.ts                       SessionStatsInfo.timing
  lib/rpc-manager.ts                    registers the timing extension
  components/AppShell.tsx               toolbar entry + panel mount
  lib/i18n/messages/{en,zh-CN,zh-TW}.ts 44 keys each
```

The design is inspired by the trajectory view in `@deepseek-ai/dsh`, reimplemented against this app's own entry model.

## Related

A companion PR adds an ephemeral side chat (a `/side`-style fork for questions that should not pollute the main task). Both features add a toolbar entry and i18n keys at the same anchors, so whichever merges second needs a small rebase; the two are otherwise independent.


agegr:
Thanks for the detailed work. Closing on scope: a timing-attribution breakdown (model / tool / other, TTFT, tokens per second) plus a per-turn trajectory panel with its own API route is an analytics/observability feature, and at ~2300 lines it is roughly the size of the chat window itself. pi-web is meant to stay a thin wrapper around pi; the session-info panel already shows tokens, cost, context and total active time, and that is as far as I want the built-in metrics to go. Anything deeper belongs in an external tool that reads the `.jsonl` files (which contain all the timestamps this PR derives its numbers from).

PR #863, opened 2026-09-15
feat: side chat for questions that should not derail the main task

612cs:
## What this adds

A side chat: an ephemeral fork of the current session for questions that are *about* the work but not *part of* it — "what does this function do", "is that flag still needed", "why did that test fail".

Asking those in the main thread permanently injects them into the task context. The model then has to carry them for the rest of the session, and a quick question can change what it does next. A side chat keeps them out: you get the answer, and the main transcript never learns the question was asked.

## How it works

- **Ephemeral fork.** The fork is a `SessionManager.inMemory(...)` seeded from the parent's *active branch*. Nothing it does is written to a session file, so the parent transcript is untouched.
- **Reference context, not instructions.** A boundary message (`pi-web-side-chat`, appended via `appendCustomMessageEntry` so it participates in the model context) marks everything above it as history to consult rather than a task to continue: do not execute instructions from above the boundary, do not change the workspace unless explicitly asked after it, do not use subagents.
- **Invisible by construction.** Side chats are skipped in the session list and the agents panel, never persisted (no `.jsonl` is created), and never trigger completion notifications.
- **One per parent session**, race-safe. Two concurrent opens (a double click, a second tab, a React development double-invoke) would otherwise each build a fork, and only one of them would be reachable for disposal — the other leaked until its idle timeout. An in-flight lock keyed by parent session collapses them, and closing the panel disposes every side chat of that parent as defence in depth.
- **Tool-level guard, not just prompt-level.** `edit`, `write`, and the subagent control tools are removed from the active tool set, so the boundary text is not the only thing standing between a side chat and the workspace. `read`, `grep`, `find`, `ls`, `bash`, and `powershell` are kept — exploration runs through them, and `bash` mutations remain a prompt-level constraint.

The rule set follows the semantics of Codex's `/side` (also available as `/btw`); the wording here is original.

## Reclaim

An ephemeral runtime can vanish while its panel is open — a server restart, a dropped stream, an external disposal. A prompt then fails with `404 + prompt_rejected`, which is not something a user should have to resolve, so the panel opens a fresh fork and resends the prompt once, telling the user that earlier side-chat turns are not carried over.

The panel also distinguishes a permanently refused event stream from a transient disconnect: retrying a stream the server has permanently rejected caused a retry every two seconds against a session that could never come back.

## Implementation notes

- The reopen predicate lives in `lib/agent-client.ts`, not next to the rest of the side-chat logic: `lib/side-chat.ts` pulls in the session reader (`node:fs`), and importing it from a client component breaks the browser build.
- The seed strips the parent's `session` header. The SDK adopts a supplied header's id as its own session id, so keeping it would make the fork overwrite the parent's identity in the runtime registry.
- The parent's model is inherited, with fallback allowed.

## Verification

- `npm test`: 1042 pass / 0 fail (22 added over main's 1020; new `side-chat.test.mjs` and `SideChatPanel.test.mjs`, plus two added cases in `agent-client.test.mjs`); `npx tsc --noEmit` and `npm run lint` clean.
- `npm run test:e2e`: 7/7 checks pass (against a development server, desktop 1280px and mobile 390px).
- Against a local OpenAI-compatible mock provider that **self-reports what it received**, over the real HTTP stack (23 checks):
  - `messages=5`, `roles=system,user,assistant,user,user`
  - `inherited_parent_history=yes`, `boundary_present=yes`, `side_question_present=yes`
  - `tools=read,bash,grep,find,ls`
  - the parent transcript is unchanged, no `.jsonl` is ever created for a side-chat id, side chats never appear in `/api/sessions`, and a closed side chat answers `404`
  - concurrent opens agree on one side conversation and leave exactly one runtime to dispose
- In the browser: opening, streaming an answer, "bring to main chat" (the answer lands in the main composer), closing, and self-heal after an external disposal with the notice shown.

## Files

```
lib/side-chat.ts                            seed, boundary text, tool narrowing (pure)
app/api/sessions/[id]/side-chat/route.ts    POST open/reuse, GET, DELETE
components/SideChatPanel.tsx                right-hand drawer, streaming, bring back
components/AppShell.tsx                     toolbar entry + panel mount
lib/agent-client.ts                         reclaim predicate
lib/rpc-manager.ts                          in-memory session, registry, in-flight lock
lib/i18n/messages/{en,zh-CN,zh-TW}.ts       14 keys each
```

## Known limits

- `bash` can still mutate the workspace; the guard is prompt-level there.
- Side-chat turns are not preserved across a reclaim, by design — the notice says so.
- One side chat per parent session; a second tab reuses the open one.

## Related

A companion PR adds session timing attribution and an event trajectory panel. Both features add a toolbar entry and i18n keys at the same anchors, so whichever merges second needs a small rebase; the two are otherwise independent.


agegr:
Thanks — the motivation is real, but I'm closing this on scope. A side chat is a new session concept (ephemeral in-memory fork, a boundary message injected into model context, a dedicated panel and API route, special-casing in the session list and agents panel) and at ~1400 lines it adds a parallel conversation model next to the existing one. pi-web already has the primitive for "ask about the work without polluting the task": **fork** the session from the current message, ask there, and delete or ignore the fork. That keeps one session model and lets the user decide whether the side conversation is worth keeping.

PR #867, opened 2026-09-16
feat: show provider quota beside the model selector

zhangnan:
## What

The composer toolbar has a flexible gap between the model selector and the right-hand controls. This fills it with a compact allowance chip for the selected provider:

`[attach] DeepSeek V4.1 Flash   Go 2/18/35%   auto   default`

usage chip

Hovering the chip lists every window with its reset time.

## Why not the existing extension status shelf

`ctx.ui.setStatus()` already reaches the shelf below the composer, but Pi extensions run *inside* a session, and a brand-new composer has no RPC session yet — so the shelf is empty exactly when a user has just picked a model and would most like to see the remaining allowance. This is what motivated the change: on an empty composer nothing renders at all.

Pi Web's provider usage adapters do not have that constraint — `queryProviderUsage()` resolves credentials through the model runtime — so this reuses that path. The chip renders with no session, and the other adapters get the same treatment for free.

## Changes

**Adapters**

- `lib/provider-usage-ids.ts`: add `opencode-go`.
- `lib/provider-usage.ts`: query `GET https://opencode.ai/zen/go/v1/usage`, which reports the *used* percent for three nested windows of one monthly allowance (5-hour / weekly / monthly, per <https://opencode.ai/docs/go/>). The normalizer maps that onto the existing percent buckets and records a non-`ok` window status in `notes`.

**UI**

- `components/ModelUsageChip.tsx` (new): paints from the shared `pi-web:provider-usage:` snapshot so a provider switch renders instantly, then refreshes on provider change, on each settled turn, on tab focus, and on a 60 s poll. It stays silent on query failure — the Models panel already owns usage error reporting.
- `components/ChatInput.tsx`, `components/ChatWindow.tsx`: render the chip in the toolbar gap. `ChatWindow` passes the message count as the per-turn refresh key, so the numbers follow the conversation.
- `app/globals.css`: `.model-usage-chip`.
- i18n: three new `providerUsage.*` keys across `en`, `zh-CN`, and `zh-TW`.

## Tests

`npx tsc --noEmit`, `npm run lint`, and `npm test` all pass — 1034 tests, 7 of them new:

- adapter: percent mapping including reset times, non-`ok` status becomes a note, missing windows throw.
- chip: compact percentage format, mixed-unit labelling, tooltip lines, provider notes.

## Notes

- The chip only renders for providers that have a usage adapter, so selecting any other provider issues no request.
- `POLL_INTERVAL_MS` is the only tunable; 60 s felt right because allowance windows move on the scale of a turn rather than a second.
- Server-provided bucket labels and window states render untranslated, matching the existing `ProviderUsageSummary`.


agegr:
Thanks. Two parts here, handled differently:

- The **OpenCode Go adapter** (`opencode-go` in `provider-usage-ids.ts` / `provider-usage.ts`) is the same change as #844, which I've approved — so that half is covered.
- The **usage chip in the composer toolbar** I'm declining. Provider quotas already have a home in Settings → Models (`ProviderUsageSummary`), and pi-web tries to show each piece of information in one place. A chip next to the model selector adds a second surface for the same data plus a network request on every model switch in every composer, and the toolbar is the one strip that needs to stay quiet. The "empty composer has no session so the extension shelf is empty" argument is fair, but it argues for looking in Settings before starting, not for a permanent indicator in the input row.

Closing in favor of #844.

PR #868, opened 2026-09-16
feat(plugins): show package description in the Plugins panel

bmdy5:
## Summary

The Plugins panel reads each installed package's `package.json` for its name and version, but drops the `description` that published pi packages declare. The list shows only the source spec — `npm:pi-cache-optimizer`, `git:github.com/DietrichGebert/ponytail` — so the panel answers "what is installed" but not "what does it do", even though the answer is already on disk in the exact file the route opens.

This surfaces it in the package detail view.

## Changes

| File | Change |
| --- | --- |
| `app/api/plugins/route.ts` | `readPackageMetadata()` also reads `description` (string-checked like `name` / `version`); `readPlugins()` returns it on `PluginPackageInfo` |
| `lib/api-types.ts` | `PluginPackageInfo.description?: string` |
| `components/PluginsConfig.tsx` | renders it as the first detail row, above Status; the sidebar item `title` falls back to it |
| `lib/i18n/messages/{en,zh-CN,zh-TW}.ts` | new `i18n.description` key |

## Notes

- Nothing new is fetched — the value comes from the local `package.json` that `/api/plugins` already reads for the version.
- It is the first grid row because it is the one field that explains what the package *is*; Status → CWD keep their existing order.
- A package with no `description` renders exactly as before (no empty row, no placeholder), and a non-string `description` is ignored just like `name` / `version`.
- The sidebar row is a fixed 30px single line shared by the Models, Skills and Sub-agents panels, so it still shows the source spec and the description is its tooltip instead. Rendering it as a second line would mean changing that shared component — happy to do it as a follow-up if you want it.

## Screenshots

`npm run dev` against a real global package set:

**English**

Plugins panel with the description row

**简体中文**

插件面板的「描述」行

## Verification

```
node_modules/.bin/tsc --noEmit   # clean
npm run lint                     # clean
npm test                         # 1028 pass, 0 fail
```

The new case in `app/api/plugins/route.test.mjs` installs a local package whose `package.json` declares a description and asserts it comes back on the API response. It fails with `actual: undefined` if the route change is reverted.


agegr (approved):
Reads a field the route already opens the file for, string-checked like `name`/`version`, renders nothing when absent. Small and useful. Merges cleanly; tsc, eslint, and the plugins route test pass. Approving.

PR #870, opened 2026-09-16
feat: add a toggle for hidden extension messages

syansunyang-collab:
## Summary

Respect `display: false` for extension messages in the chat view by default, with an opt-in setting to show them.

Extensions can persist hidden `custom_message` entries for model context. Rendering each entry as a collapsed card can crowd the conversation with messages that the extension explicitly marked as hidden.

## Changes

- Add **Show hidden extension messages** under **Settings → General → Chat**, off by default.
- Persist the preference in localStorage alongside existing chat appearance settings.
- Filter only `role === "custom" && display === false` in the client-side display projection.
- Keep message entry IDs aligned with the filtered messages.
- Preserve visible extension messages and messages without an explicit `display: false` flag.
- Add English, Simplified Chinese and Traditional Chinese labels.

Session files, extension execution, summaries sent to the model and tool-result retention are unchanged.

## Tests

- Unit coverage for filtering, entry ID alignment, source immutability and opt-in passthrough.
- Settings layout coverage updated for the new switch.
- Browser coverage with 13 hidden extension messages: hidden by default, shown on opt-in, persisted after reload, and hidden again on opt-out; regular messages remain visible.

## Validation status

- Targeted unit/contract tests: passed (26 tests).
- TypeScript and changed-file ESLint: passed.
- Production build: passed.
- The new hidden-message assertions passed in a production-mode Microsoft Edge run on Windows.
- Upstream CI `checks` passed for this PR: https://github.com/agegr/pi-web/actions/runs/35050124466
- The full browser suite stops at `Keyboard selection must scroll into view` in `e2e/extension-dialog.mjs:51`. The exact same assertion already fails on the unmodified base commit `fcd94bf` in https://github.com/agegr/pi-web/actions/runs/34967181959 . The dialog code/test is not changed by this PR.
- A focused Windows/Edge run excluding that separate dialog suite passes the new toggle/persistence assertions, but its final browser-error guard also reports HTTP 500 for the compacted-session fixture state endpoint. That local fixture failure remains disclosed and has not been independently attributed. No existing assertions have been removed or suppressed in the submitted changes.


syansunyang-collab:
Note on the red `e2e` check: it fails on `Keyboard selection must scroll into view` (`e2e/extension-dialog.mjs:51`), which is unrelated to this change. The same assertion fails on `main` at this PR's merge base `fcd94bf` — see the push run https://github.com/agegr/pi-web/actions/runs/34967181959 (job 104374316455): same two `PASS:` lines, then the same `AssertionError`. This branch is one commit on top of that base and the `checks` job passes. Happy to rebase and re-run once `main` is green.

agegr:
Thanks for the careful implementation and the note about the unrelated e2e failure. I'm closing this one, though.

Rendering `display: false` extension messages as a collapsed, labelled card (rather than dropping them like the TUI does) is a deliberate choice in pi-web: it keeps what extensions inject into the model context visible to the person reading the transcript, which matters when debugging an extension or understanding why the model behaved a certain way. Adding a Settings toggle to hide them makes that a per-browser preference and grows the Settings surface, which pi-web tries hard to avoid; a single good default is preferred over a switch.

If the concrete problem is a run of many consecutive hidden messages crowding the view, the acceptable shape would be visual — e.g. collapsing a consecutive run into one row that expands — with no setting. Happy to look at that as a separate PR.

PR #872, opened 2026-09-16
feat: add a compact tree view for changed files

Kitten9533:
## Problem

The explorer changes pane only showed a flat list of git files. In a project with nested paths that is hard to scan. VS Code Source Control has **View as List** and **View as Tree**; this change adds the tree option without replacing the list.

## Solution

The changes header (`N files +x -y`) gets a list/tree toggle on the right.

**Tree**

- Shows only changed files, nested by directory.
- Single-child folders are compacted (`src/components/Button.tsx` → `src/components`).
- Switching from list to tree expands every folder. Collapsing a folder is kept if you close and reopen the changes pane while still in tree mode.
- Folders get a translucent status dot on the right. One kind of change uses that color (add/untracked green, modified yellow, deleted/conflict red, renamed blue). Mixed kinds use yellow; any conflict uses red.
- Files keep the same `M` / `A` / `D` / `R` / `U` / `C` letters as the list, on the right.
- Deleted file names are struck through, still using the normal text color.

**List**

- Same files as before.
- Status letters move to the right, matching the tree and VS Code SCM.

The chosen view is stored in `localStorage` (`pi-web:file-explorer:changes-view`). Default remains list.

Clicking a file still opens it with `modeHint: "diff"`, using the original `GitFileStatus.filePath` so the tree does not rebuild a path by joining `cwd` + relative path.

## Changes

- `components/FileExplorer.tsx` — list/tree toggle, compact tree, folder dots, strikethrough, open-by-original-path
- `lib/search-tree.ts` — `compactFolders`
- `lib/file-explorer-state.ts` — persist view mode
- `lib/file-paths.ts` — `getChangeTreePath`
- `lib/git-status.ts` — `aggregateGitStatuses`
- `lib/i18n/messages/{en,zh-CN,zh-TW}.ts` — View as Tree / View as List

## Tests

```
node --experimental-strip-types --test lib/search-tree.test.mjs lib/file-explorer-state.test.mjs lib/git-status.test.mjs lib/file-paths.test.mjs components/FileExplorer.test.mjs
```

`npm test`: 1039/1039 pass. `tsc --noEmit` and eslint on the touched files are clean.

---

## 问题

文件浏览器的变更列表以前只有扁平路径。目录一深就不容易扫。VS Code 源代码管理有列表和树两种看法，这次补上树状看法，列表还在。

## 解决方案

变更标题栏右侧可以在列表和树之间切换。

- 树只显示变更文件，按目录嵌套，单子目录会压成一行。
- 从列表切到树时文件夹全部展开。面板开着时折叠仍有效；关掉再打开会记住折叠，不会强制全展开。
- 文件夹右侧有半透明状态点：单一操作用对应颜色，混用黄色，有冲突用红色。
- 文件状态字母（M/A/D 等）在列表和树里都在右侧。
- 删除的文件名加删除线，字色不变。
- 列表/树选择会记住，默认仍是列表。
- 点文件仍打开 diff，路径用 git 返回的原始路径。

## 测试

相关测试与 `npm test` 均通过。

https://github.com/user-attachments/assets/8966181f-4b3a-4a73-8a44-b3c319e09dcf


agegr:
Thanks — this is well built (path compaction, status roll-up, persisted view mode, tests), but I'm closing it on scope. The explorer's changes pane is meant to be a quick flat list of what git reports; a second, VS Code-style tree mode with its own toggle, expansion state, folder status dots, and localStorage key is ~500 lines of IDE chrome for the file panel, and it would need to be kept in step with every later change to the explorer (#853 just added a mention button to the same rows, for instance). pi-web tries to stay a thin wrapper around pi rather than grow into an editor.

The one small piece I'd take on its own is moving the status letters to the right edge of the list rows, if you want to open that separately.

Issue #764, opened 2026-09-08
Feature Request: Allow deleting / hiding project workspaces from the sidebar


zhubaodian1027:
### Problem Description

Currently, the sidebar groups sessions by project (`cwd`). Users can easily add a new project workspace via the directory picker when creating a new session, but there is no way to remove or delete a project from the sidebar once it has been added.

To remove a project, users currently have to either:
1. Manually delete every single session inside the project one by one through the UI.
2. Go to `~/.pi/agent/sessions/` and manually `rm -rf` the encoded project directory.

This makes workspace management cumbersome, especially when dealing with temporary projects, test directories, or outdated projects.

### Proposed Solutions

Any of the following (or a combination) would greatly improve the UX:

1. **Delete Project (Batch Delete Sessions)**
   - Add a context menu or a "Delete" button (trash icon) on project workspace hover.
   - Show a confirmation dialog: *"This will delete all N sessions under this project. Project files on disk will not be affected."*
   - Behind the scenes, delete all session `.jsonl` files (or the project session folder) and invalidate the session list cache.

2. **Hide / Archive Project (Soft Removal)**
   - Allow hiding a project from the sidebar without deleting session history on disk, perhaps persisted in local storage or a config file.

3. **Empty Project Cleanup**
   - Automatically clean up / hide projects that have no active sessions.

### Environment
- `@agegr/pi-web`: 0.8.11+
- OS: macOS

Issue #642, opened 2026-08-28
Feature request: remove/close stale workspace entries from the workspace picker (still absent in v0.8.11; relates to #342)


like-attract:
## Problem

The workspace list is derived from on-disk session folders / session `cwd` values and only ever grows — there is no way to remove or "close" a workspace entry from the picker. I checked and there is no button for it in the current UI (v0.8.11, the latest release).

This becomes painful in a very common scenario: **renaming or moving a project directory**.

1. I had sessions recorded under `D:\...\26投递\简历`.
2. I reorganized and the project now lives at `D:\...\26投递` (folder renamed / sessions re-homed).
3. Result: the old `简历` workspace entry stays in the picker forever and cannot be dismissed, while the new workspace shows "no sessions" until the stored `cwd` in the session files is manually rewritten.

So a one-off or renamed directory leaves a permanent orphan entry, and the list keeps getting longer — exactly the pain described in #342.

## Relation to #342

#342 ("最近工作区下拉列表支持移除过期条目") requested the same capability and even linked a reference implementation (`feat/cwd-delete`, commit `d41fd17`), but it was **closed without being merged**. Confirming the feature is still absent in **v0.8.11** (latest). This issue re-raises it with a concrete rename/orphan reproduction on the latest version, and is not meant to duplicate an open request.

## Expected behavior

- Each **non-current** workspace entry exposes a remove control (e.g. an × at the row end); clicking removes it from the picker.
- Removal affects **display only** — it must not delete sessions or the on-disk directory.
- Removal state is persisted (localStorage is fine).
- If the directory is later re-selected (Custom path / default), the entry reappears, so a mistaken removal is recoverable.

## Suggested minimum acceptance criteria

- [ ] A stale/renamed workspace can be removed from the picker without touching its sessions or files.
- [ ] Removed entries stay hidden across restarts.
- [ ] Re-selecting the directory restores its entry.

Happy to test a patch or adapt the reference branch from #342 if useful.

agegr:
The current directory list is automatically generated by the session and is not an independently stored historical record. After clearing sessions in that directory, the directory will automatically disappear. Shift + Delete can skip confirmation; When there are many table of contents, filters can also be used. Therefore, no separate cleanup button will be added for now, to avoid confusion with batch deletion sessions. If there is a clear need to retain sessions but hide directories later, consider adding archiving capabilities.

PR #876, opened 2026-09-16
feat(sidebar): hide projects from the picker without deleting sessions

ani13-teach:
Implements the **hide / archive half of #764** — the part left open in the #642 close note (*"no separate cleanup button … if there is a clear need to retain sessions but hide directories later, consider adding archiving capabilities"*).

So, to be explicit about scope: **nothing in this PR deletes sessions, and there is no cleanup/batch-delete button.** Only the picker row goes away, and it can be brought back with one click.

## The problem this addresses

The picker lists every directory that has session history, derived from the sessions themselves — there is no stored registry to edit. That makes a stale entry expensive to remove: the only available lever was deleting the project's sessions, so a single visit to a temporary directory cost history that had nothing to do with the clutter. #764 documents the same two workarounds (delete sessions one by one, or `rm -rf` the encoded directory under `~/.pi/agent/sessions/`).

## What it does

- Every picker row gets a hide control next to it: a **crossed-out eye, deliberately not a trash icon**, because the sessions are kept. It stays visible rather than appearing on hover, since hover does not exist on the mobile layout this app supports.
- The picker footer shows `Hidden projects (N)`, which expands to the hidden paths with a one-click restore. No confirmation step on either side — the action is reversible, unlike deleting a session or a worktree, and that is the difference the #642 note asked for.
- Hidden projects are also skipped by the picker's initial auto-selection, so a project you hid is not the one that gets reopened on load.

## What deliberately does not change

- `GET /api/sessions` still returns every session and no `.jsonl` file is touched — hiding is purely a picker filter.
- Hidden entries key off the same stable project identity the list already groups by (`projectKey`, case- and separator-insensitive on Windows), so hiding one repository hides all of its worktrees and path spellings in one go, and the entry stays restorable for as long as that project still has sessions.

## Storage choice

`localStorage` under `pi-web:hidden-projects`, alongside the other view preferences (workspace memory, tool presets, chat appearance). I picked it over a config file because it keeps the server path and `~/.pi/agent` untouched for what is a view preference.

The trade-off is real, though: a project hidden on the desktop is still visible on a phone or another browser. If you would rather it be server-side, it is a small change at the same two call sites — I did not want to add server state without asking. Happy to add a `docs/adr` note for the choice as well.

## Verification

`npx tsc --noEmit`, `npm run lint` and `npm test` all pass. New coverage:

- `lib/hidden-projects.test.mjs` — storage round-trip, corrupt/wrong-shape/junk values, empty list clearing the key, and blocked or throwing `localStorage`.
- `partitionRecentProjects` in `lib/project-groups.test.mjs` — row split, and all worktrees of one repository hiding together.
- `components/SessionSidebar.test.mjs` — source assertions that the hide handler never reaches the network or a `DELETE`, which is the whole point of the feature.

I also drove it in a real browser against an isolated `PI_CODING_AGENT_DIR` fixture with three fake projects: hide → two rows plus a footer count and one persisted key, reload keeps it hidden, restore returns the row and clears the key, `/api/sessions` still returns all three sessions, and the fixture `.jsonl` files are byte-identical before and after. Happy to turn that into an `e2e/` check for `npm run test:e2e`, or to post the screenshots, if you want either in the repo.

Refs #764


agegr:
Thanks for the well-scoped proposal and for making the non-destructive nature explicit. I'm closing this for now for the same reason as #824 and #834: session/project organization in the sidebar is about to be redesigned, and hiding/archiving projects is squarely part of that design rather than something I want to add to the current picker first. The #642 note you quote still stands — archiving is on the list — but it will land as part of the redesign so the model (what is hidden, where it is stored, how it interacts with grouping) is decided once. I'll keep this PR in mind when that work starts.

PR #879, opened 2026-09-16
fix(pwa): bound navigation and asset fetches so a dead upstream cannot hang

YsDirector:
## Summary

Follow-up to #805, scoped to the one part you asked to land on its own.

A reachable port with a dead upstream accepts the connection and then never answers, so `fetch()` neither resolves nor rejects. The navigation never reached its `.catch`, so it never showed `offline.html` — the page stayed blank until the browser gave up. A static asset cache miss had the same problem.

This bounds both network waits. No shell or data caching is added, so the caching surface is unchanged.

- navigation: `fetchWithTimeout(request, NAVIGATION_TIMEOUT_MS)` (8s) in front of the existing `offline.html` fallback
- assets: the cache-miss `fetch()` in `cacheFirst()` gets the same ceiling
- the budget covers time to first byte only

## Why the timer is cleared when the headers arrive

`fetchWithTimeout()` clears its timer as soon as the response headers land, so the ceiling applies to the wait for a first byte rather than to the whole response. Next.js streams its SSR payload, and a body that keeps streaming past 8s must not be aborted mid-stream. There is a test for exactly that.

## Behavior

| URL | Dead upstream | Notes |
|---|---|---|
| navigation | `offline.html` after 8s | before: hung forever |
| `/_next/static/*`, cache hit | cache | unchanged |
| `/_next/static/*`, cache miss | bounded rejection after 8s | before: hung forever. A rejected asset fetch is already what happens when the network is down |
| `/api/*` | untouched | unchanged |

## Testing

```
node --test public/sw.test.mjs   # 8 passed
npm test                         # 1038 passed
npm run lint                     # clean
```

Three new tests use `node:test`'s fake clock: a stalled navigation falls back to `offline.html`, a stalled asset request is bounded, and a response that arrives in time is not aborted afterwards. Reverting only `public/sw.js` makes exactly those three fail, so they are not vacuous.

## Notes

8s is a pick, not a measurement — happy to change it to 3s or 5s. I kept the two constants separate so navigation and assets can be tuned independently.

If you would rather not take the asset half, dropping the one line in `cacheFirst()` and its test leaves the navigation fix self-contained.


agegr (approved):
Exactly the piece I asked for in #805, and scoped correctly: an `AbortController` ceiling on time-to-first-byte for navigations and asset cache misses, timer cleared once headers arrive so streamed SSR bodies are never cut, no shell caching, `/api/*` untouched. Tests pass (8/8), lint clean, merges cleanly. Approving — thanks for the quick follow-up.

